PR #5, opened 2014-07-16
added glyph-outline-standalone example

vishnubob:
Hi, thank you for this great library!  I'm using it to add font capability to my OpenSCAD python package.  I needed to rasterize the outline, so I hand rolled my own bezier curve function.  I figured this script would be helpful for anyone trying to do the same thing.  Since it only depends on python's standard library, it could be easily integrated with your library.


rougier:
Any reason why this wasn't merged ?

PR #20, opened 2015-01-13
Split the ctypes function binding out into raw.py.

mdsitton:
This should make things a bit better for people who want to use the ctypes binding directly, instead of the pythonic wrapper.

This still may need some work, and i haven't tested it thoroughly.

### DIFF
--- a/freetype/__init__.py
+++ b/freetype/__init__.py
@@ -16,54 +16,23 @@ Note: C Library will be searched using the ctypes.util.find_library. However,
       this search might fail. In such a case (or for other reasons), you may
       have to specify an explicit path below.
 '''
-import os
 import sys
-import platform
 from ctypes import *
-from freetype.ft_types import *
-from freetype.ft_enums import *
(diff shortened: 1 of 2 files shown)


rougier:
Good idea but it could be better to have the raw stuff in the__init__ and the pythonic api in another file, no ?

mdsitton:
Possibly, but my reasoning is as follows:
1. Keep compatibility with all current code
2. Allow one to use only the raw api and not have to pull in the pythonic api.

rougier:
Make sense. I'll test it and I'll merge (if tests are ok of course)

mdsitton:
I do think it might be a good idea to logically split up, and move stuff out of __init__.py for maintainability reasons.

mdsitton:
Any updates on merging?

rougier:
Sorry, I forgot... Just merge it. Thanks.

mdsitton:
No problem.

PR #21, opened 2015-01-25
Fix FT_Done_FreeType never being called on FT_Library instance.

mdsitton:
In python 2 the injected \_\_del\_\_ method was called, but by the time FT_Library was deleted the FT\_Done\_FreeType function was gone. This has been worked around by storing a reference of FT\_Done\_FreeType.

However under python 3  there was a second issue. For whatever reason the injected \_\_del\_\_ method wasn't called at all. They work on a normal classes and subclasses in python 3, so I'm not completely sure what was going on there. Fixed this by just subclassing FT_Library, and defining the needed stuff in there.

This also has a nice side effect of no longer injecting the \_\_del\_\_ function into the raw FT_Library, which also effected anyone using freetype raw.

### DIFF
--- a/freetype/__init__.py
+++ b/freetype/__init__.py
@@ -24,7 +24,7 @@ from freetype.raw import *
 PY3 = sys.version_info[0] == 3
 if PY3: unicode = str
 
-__handle__ = None
+_handle = None
 
 
 FT_Library_filename = filename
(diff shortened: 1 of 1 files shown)


freetype/__init__.py
```diff
@@ -24,40 +24,50 @@
...
+def _init_freetype():
+    global _handle
+
+    _handle = _FT_Library_Wrapper()
+    error = FT_Init_FreeType( byref(_handle) )
+
+    if error: raise FT_Exception(error)
+
+    try:
+        set_lcd_filter( FT_LCD_FILTER_DEFAULT )
```
rougier: While reading your patch, I wonder why I put this specific initialisation code here. In any case, we should not silence the exception.

mdsitton:
I agree, also set_lcd_filter It also happens eventually to loop back and call get_handle before _init_freetype() returns.

mdsitton:
It was added in this commit: 2fc263659922ea2302afc498f3592938482c5c86

mdsitton:
Any status on merging?

rougier:
Sorry for the delay.

PR #24, opened 2015-02-15
Binding rework

mdsitton:
I was having some annoyances when using the raw api by itself some hard to fix bugs because of python segmentation faults. So i decided to go through and redo the ctypes binding a bit. Things should be a lot less fragile when using it standalone now.

There are still a couple of thing i need to do. As it stands if you try to use an older version of freetype you will most likely get an error on import because of it unable to find a newer function. I want to get that fixed before this is merged.

Eventually I'd like to go through and fill out more of the freetype api than what is defined so far just so things are more complete. For that i went and put comments of the names that freetype gives its headers. Then for headers that already had some functions defined from it i went and put a comment with the names of functions not yet defined.

Also for the binding function i could use the restype and argtypes method, but i just generally prefer to do it the way i did in the code if I'm writing a function that does the stuff.

So in short don't quite merge yet one thing i still need to do, just putting the pull request up for any comments for now.

### DIFF
--- a/freetype/__init__.py
+++ b/freetype/__init__.py
@@ -27,8 +27,6 @@ if PY3: unicode = str
 _handle = None
 
 
-FT_Library_filename = filename
-
 class _FT_Library_Wrapper(FT_Library):
     '''Subclass of FT_Library to help with calling FT_Done_FreeType'''
     # for some reason this doesn't get carried over and ctypes complains
--- a/freetype/ft_enums/__init__.py
+++ b/freetype/ft_enums/__init__.py
@@ -109,6 +109,7 @@ from freetype.ft_enums.ft_render_modes import *
 from freetype.ft_enums.ft_stroker_borders import *
 from freetype.ft_enums.ft_stroker_linecaps import *
 from freetype.ft_enums.ft_stroker_linejoins import *
+from freetype.ft_enums.ft_size_request import *
 from freetype.ft_enums.ft_style_flags import *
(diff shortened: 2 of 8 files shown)


rougier:
We may need to extract the actual installed freetype version and use your def_func to raise a warning if the function is not available. And make the freetype version available.

I'm also thinking of setting the freetype-py version to match the latest version of freetype so people know when to update. What do you think ?

mdsitton:
Thats what i was mostly planning to do i'm not sure if getting the freetype version will be required though. Sounds like a good idea. However I would wait until all of the functions for that version are bound before the version is changed.

mdsitton:
While doing this i have thought of a few other things i *might* want to do in this branch. I may redo how enums are defined. Need to think a bit more about how though, I do have a couple interesting ideas though.

mdsitton:
So this should work in as a method of binding the enum's but i haven't tested it with ctypes yet.
https://gist.github.com/mdsitton/0cdafcca735c21c7b7e8

rougier:
Can you give a summary of your changes ?
We might need to update the README and the doc accordingly.

mdsitton:
Yeah i can, ill get that ready after i'm finished with the enums.

mdsitton:
Sorry that there hasn't been much activity on this, i've been a bit busy for a while. Might be able to get back to this here soon.

rougier:
No problem.

rougier:
Closing until further notice from @mdsitton

PR #29, opened 2015-06-03
better loader

Tcll:
support for sys.path.append('custom directory/') before importing, and better version (should support pygame's freetype library) and error handling. (though it could all be better)


Tcll:
did something even better so I'm submitting a new one :)

PR #33, opened 2015-12-26
Face: allow creating a font from memory

### LINKED ISSUE #32: Possible to create a Font object from memory?
A binding to `FT_New_Memory_Face` for creating a Font object. I have a binary font in memory and I don't want to save it to disk and load it again.

Is it possible to do so? 
The Font constructor in the docs seems to only take a filename.

adrientetar:
~~Unresolved question: I made the Face object always read the file object using Python internally while previously it would try to pass the filename to FT functions before reading the file using Python. 
Maybe that's undesired for some performance/memory reasons(?) I really don't know.~~

Anyway. Closes #32.

### DIFF
--- a/freetype/__init__.py
+++ b/freetype/__init__.py
@@ -964,19 +964,28 @@ class Face( object ):
                The index of the face within the font.
                The first face has index 0.
         '''
-        with open(filename, mode='rb') as f:
-            filebody = f.read()
-        self._init(filebody, index)
+        self._init(filename, index)
 
-    def _init(self, filebody, index):
+    def _init(self, fileOrData, index, isData=False):
         library = get_handle( )
         face = FT_Face( )
         self._FT_Face = None
-        #error = FT_New_Face( library, filename, 0, byref(face) )
         self._filebodys = []
-        error = FT_New_Memory_Face( library, filebody, len(filebody),
-                                    index, byref(face) )
-        self._filebodys.append(filebody)  # prevent gc
+        if not isData:
+            # fileOrData is filename
+            try:
+                u_filename = c_char_p(_encode_filename(fileOrData))
+                error = FT_New_Face( library, u_filename, index, byref(face) )
+            except UnicodeError:
+                # read filename and switch to filebody
+                with open(fileOrData, mode='rb') as f:
+                    fileOrData = f.read()
+                isData = True
+        if isData:
+            # fileOrData is filebody
+            error = FT_New_Memory_Face( library, fileOrData, len(fileOrData),
+                                        index, byref(face) )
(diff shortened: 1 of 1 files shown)


rougier:
Thanks for the PR !

I think it would be better to have FT to read the filename since it will also give us a relevant error message.

Also, what do you think of having:

```
def __init__( self, filename=None, index = 0, buffer=None ):
```

This would not change the current API but would allow to have the buffer kayword meaning the font file has been already loaded.

adrientetar:
Okay I put an internal dispatch so it's DRY but uses relevant FT APIs.

adrientetar:
Btw I like my updated solution better than what you proposed because essentially the buffer argument that you suggest is a substitute for filename &ndash; so it's why I went down making a separate constructor (it's clearer and more pythonic to do so imho).

rougier:
Thanks. Then it would be better to have both a `_init_from_file` and `_init_from_memory` and the `__init__` method would decide which one to call based on `data` type. And we may add a `from_file` class method alongside your `from_memory`. I can merge your PR and do the change if you prefer.

adrientetar:
Hmm... it looks to me like this makes things more complicated than they otherwise need be, can you detail what you dislike the current patch with alternate constructors? The `data` variable you suggest would be a string regardless of which constructor you want to call afaict so as detailed e.g. [here](http://stackoverflow.com/a/141777/2037879) I think it makes all the more sense to do it like this.

Given my patch you can either do `face = Face("./Vera.ttf")` or `face = Face.from_memory(contents)` as you would do in C++ for instance (with a static method, I mean).

The current code has chosen to put filename in the default constructor, and probably rightfully so since it's likely how most people use it, so another advantage of an alternate constructor is to keep current code paths as they are.

To be clear I'm not against changing things per se. but I just wish to understand why you say it would be better to do as you suggest.

Thanks!

rougier:
The code would be approximately the same but it would be more clear. I don't like the __init__ method that is now just a call to the _init. I would prefer the __init__ to be the place where the test against file_or_data is made and the proper method to be called. This is only  a matter of reorganising your code a little bit.

HinTak:
strings are not the same as byte buffers. The distinction is clearer, especially in python 3.x . Also, you will need to worry about whether the user does `from __future__ import unicode_literals` . Anyway, I am all for the constructor auto-select based on the input type.

olivierberten:
Any reason why this hasn't been merged?

rougier:
Sorry, I think I just forgot. This probably needs a rebase at least (@adrientetar ?)

yifeikong:
Wouldn't it be better to read from a file-like object? the logic would be simple. besides If you do need to load from a bytes, just use io.BytesIO

Issue #49, opened 2017-03-28
Freetype API - load flags - not quite up to date.


HinTak:
I found `FT_LOAD_COMPUTE_METRICS` not yet in freetype-py . It was introduced with freetype 2.6.1 . ( I contributed that change upstream, and also pushed the equivalent to the C# binding in SharpFont a while ago).

Also a related question: what version of freetype is freetype-py last sync to? Or is it just added as needed?

HinTak:
Another API, FT_get_property/FT_set_property (introduced at the end of 2.6.x or beginning of 2.7.x) is also missing.

HinTak:
There is an example to

HinTak:
There is part of an example in my fontval-diag branch on how to use property_set/get

rougier:
There is not official freetype version but now it might be a good idea to track freetype version such that people knows what version they need. We could target a freetype-py 2.6 release for a start, what do you think ?

HinTak:
Some of the LTS release of ubuntu or redhat enterprise linux may be shipping things as old as 2.4.x . OTOH, in my other comment ( https://github.com/rougier/freetype-py/pull/50#issuecomment-290502221 ) I see freetype-py currently only uses 65 of some 190+ exported routines from freetype, so there is definitely plenty of room for improvement.

One thing that worries me a bit is that freetype-py looks for all the freetype routines at load time (unless try:
except AttributeError: ) . This means if the found libfreetype does not contain every one referenced, even when the user has no intention of using the more obscure ones, trying to use freetype-py for anything at all would fail. Is there anyway ctypes can do late-binding? Otherwise, putting a lot of try/except around 120+routines is going to be painful.

BTW, freetype's export symbols or what not, had entered linux-standard-base ( http://www.linuxfoundation.org/collaborate/workgroups/lsb ) a while ago, so that's probably one starting point to know what popular/minimal freetype versions are out there.

HinTak:
I downloaded a few past versions from fedora's build farm ( https://koji.fedoraproject.org/koji/packageinfo?packageID=313 ) and extracted the symbol lists. Strangely enough the smallest number of exported routine is at 190 for 2.6.5 . The difference with 2.7.X seems reasonable - 3 more on multimaster/variation fonts. A bunch of symbols on the `FT_Stream_*` has gone missing between 2.6.0 and 2.6.5 (or maybe build system difference also). The oldest I got, 2.2.1 still shares about 160 routines with current.

[...]

PR #50, opened 2017-03-28
Adding two new load flags, and their associated documentation from up…

HinTak:
…stream.

FT_LOAD_COMPUTE_METRICS was introduced upstream in FreeType 2.6.1 by me,
as part of the enhancement for the new Font Validator backend.

Also re-indent the previous's documentation, FT_LOAD_COLOR, like the rest.

Part of the fix to
https://github.com/rougier/freetype-py/issues/49

### DIFF
--- a/freetype/ft_enums/ft_load_flags.py
+++ b/freetype/ft_enums/ft_load_flags.py
@@ -137,11 +137,28 @@ FT_LOAD_NO_AUTOHINT
 
 FT_LOAD_COLOR
 
- This flag is used to request loading of color embedded-bitmap images. The 
- resulting color bitmaps, if available, will have the FT_PIXEL_MODE_BGRA 
- format. When the flag is not used and color bitmaps are found, they will be
- converted to 256-level gray bitmaps transparently. Those bitmaps will be in
- the FT_PIXEL_MODE_GRAY format.
+  This flag is used to request loading of color embedded-bitmap images. The
+  resulting color bitmaps, if available, will have the FT_PIXEL_MODE_BGRA
+  format. When the flag is not used and color bitmaps are found, they will be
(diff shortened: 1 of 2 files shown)


rougier:
Thanks. Do you know how much work would it be to be fully compatible with freetype 2.6.1 ?

HinTak:
Hmm, I don't have anything as old as 2.6.1 :-). 2.7.1 exports 193 symbols, from:

```
nm -D /usr/lib64/libfreetype.so.6 | grep ' T ' | wc -l
```

whille mingw freetype 2.6.5 exports 190:
```
winedump -j export /usr/i686-w64-mingw32/sys-root/mingw/bin/libfreetype-6.dll
winedump -j export /usr/x86_64-w64-mingw32/sys-root/mingw/bin/libfreetype-6.dll
```

freetype uses gcc's symbol visibility functionality to hide internal routines, so what's exported is meant to be public and therefore these are fairly authoritative figure. AFAIC, freetype-py (from `freetype-py/freetype/raw.py` ) uses 65 currently, so there is a long way to go.

By comparison, SharpFont (C# binding for freetype) gets at 191.

```
SharpFont]$ grep DllImport Source/SharpFont/FT.Internal.cs  | wc -l
191
```

So SharpFont is more or less complete to current - although, from personal first-hand experience, much of it not tested or even been used. Also as C# is a bit further from the C-API than freetype-py, when it is not working correctly, not as obvious as freetype-py. 

BTW, freetype 2.7.2 or 2.8.0 (the next one after 2.7.1) is coming out soon in a metter of days.

rougier:
Huumm... that's a lot actually. Maybe we need to automatize things a bit then. I did it mostly by hand to get the documentation but I guess there might be some way to do the same via a python script.

PR #52, opened 2017-04-02
Adding symbols from 2.2.X onwards

HinTak:
This is taken from fedora's build farm, so it is possible that other freetype builds have fewer symbols... However, 2.2.X is more than a decade old!

### DIFF
--- a/examples/agg-trick.py
+++ b/examples/agg-trick.py
@@ -27,7 +27,10 @@ def render(filename = "Vera.ttf", hinting = (False,False), gamma = 1.5, lcd=Fals
     if lcd:
         flags |= FT_LOAD_TARGET_LCD
         Z = np.zeros( (H,W,3), dtype=np.ubyte )
-        set_lcd_filter( FT_LCD_FILTER_DEFAULT )
+        try:
+            set_lcd_filter( FT_LCD_FILTER_DEFAULT )
+        except:
+            pass
 
 
     for size in range(9,23):
--- a/examples/emoji-color.py
+++ b/examples/emoji-color.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+# -*- coding: utf-8 -*-
 
 import freetype
 import numpy as np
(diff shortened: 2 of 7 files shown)


rougier:
Thanks a lot! When you talk about builds, you mean freetype version or are there any compile options that may some symbols to be not present ?

Also, I wonder to what extent we could use context manager to restrict bindings to freetype version. Something like:

```
with version(">2.6"):
   function  = lib.function
```

but maybe it's not a good idea in the long term.

HinTak:
Many parts of FreeType are optional, or enabled by auto-detecting some dependent part. Although, it is probably rare on modern linux not to enable as many as possible. That may be all you need to care in practice.

Checking for FreeType version is not a reliable means of guaranteering a symbol being present.

rougier:
And these option are saved in `ftoption.h`, is that right ?

HinTak:
Yes, `ftoptions.h` contains the build-time configuration - however, you can't assume it exists, as it is only needed for c devel - also its location changes somewhat recently (2.5 to 2.6?).

BTW, I think my alphabetical-order dump is not ideal, it would be better by category, eventually, if some if it breaks. But it would do for now, to make as many routines available as possible.

rougier:
Looks goot to me. Should I merge or do you have more modifications in mind ?

HinTak:
All good to merge.  Please go ahead.

I tried adding a destructor to class Bitmap to call FT_Bitmap_Done to automatically clean up, but then the two gl examples misbehave. So some of the examples require manual clean up, while some forbid clean up from the python side. This is just a heads-up if you are thinking of adding a destructor: the gl examples seems to DEPEND on the python side leaking - I.e. They seem to depend on the C structures for FT_Bitmap to stay while their python wrappers go out of scope without cleaning them.

rougier:
Thanks !
I'll loon into the GL examples when I've some free time. Could you open an issue for the GL memory problem ?

HinTak:
okay, I filed https://github.com/rougier/freetype-py/issues/53 .

PR #55, opened 2017-04-26
pycairo based rewrite of the drawing examples.

HinTak:
Should work also with cairocffi.
    
    TODO/Limitations:
    
    - glyph-mono+alpha requires libtiff on small-endian platforms
    for mono-rendering to bitswap correctly. It may be possible to use
    FT_Bitmap_Convert() instead. TODO
    
**edit** I wanted to do strikethrough to keep the paragraph but indicates that the below nolonger applies.
Already fixed by further commits later.
```
   - wordle requires pycairo 1.11+ (released only on 2017-04-09)
    for the overlap check, and does not pack as tightly. Further work.
    Changing paint color/alpha also not seem to work. Investigate.
```

### DIFF
--- a/examples/bitmap_to_surface.py
+++ b/examples/bitmap_to_surface.py
@@ -14,6 +14,8 @@ Usage:
 
 Works with cairocffi too. (Replace "from cairo ..." with "from cairocffi ...")
 
+Limitation: Surface.create_for_data is not in the "python 3, pycairo < 1.11" combo.
+
 This is a heavily modified copy of a few routines from Lawrence D'Oliveiro[1],
 adjusted for freetype-py, and bugfix/workaround for mono-rendering [2].
 
--- a/examples/bitmap_to_surface.py
+++ b/examples/bitmap_to_surface.py
@@ -1,0 +1,93 @@
+'''
+FT_Bitmap to CAIRO_SURFACE_TYPE_IMAGE module
+============================================
+
+Converting from Freetype's FT_Bitmap to Cairo's CAIRO_SURFACE_TYPE_IMAGE
+
+Copyright 2017 Hin-Tak Leung
+
+FreeType is under FTL (BSD license with an advertising clause) or GPLv2+.
+Cairo is under LGPLv2 or MPLv1.1.
+
+Usage:
+     from bitmap_to_surface import make_image_surface
+
+Works with cairocffi too. (Replace "from cairo ..." with "from cairocffi ...")
+
+This is a heavily modified copy of a few routines from Lawrence D'Oliveiro[1],
+adjusted for freetype-py, and bugfix/workaround for mono-rendering [2].
+
+The bugfix/workaround requires libtiff on small-endian platforms.
+
+TODO: Look into using FreeType's FT_Bitmap_Convert() instead. However, libtiff
+      is common enough, and probably not important.
+
+[1] https://github.com/ldo/python_freetype
+    https://github.com/ldo/python_freetype_examples
+
+[2] https://github.com/ldo/python_freetype/issues/1
+    https://github.com/ldo/python_freetype_examples/issues/1
+'''
+from freetype import FT_PIXEL_MODE_MONO, FT_PIXEL_MODE_GRAY, FT_Pointer, FT_Bitmap
+
+from cairo import ImageSurface, FORMAT_A1, FORMAT_A8
+#from cairocffi import ImageSurface, FORMAT_A1, FORMAT_A8
+
+from array import array
+from ctypes import cast, memmove, CDLL, c_void_p, c_int
+from sys import byteorder
+
+def make_image_surface(bitmap, copy = True) :
+    if ( type(bitmap) == FT_Bitmap ):
+        # bare FT_Bitmap
+        content = bitmap
+    else:
+        # wrapped Bitmap instance
+        content = bitmap._FT_Bitmap
+    "creates a Cairo ImageSurface containing (a copy of) the Bitmap pixels."
+    if content.pixel_mode == FT_PIXEL_MODE_MONO :
+        cairo_format = FORMAT_A1
+    elif content.pixel_mode == FT_PIXEL_MODE_GRAY :
+        cairo_format = FORMAT_A8
+    else :
+        raise NotImplementedError("unsupported bitmap format %d" % content.pixel_mode)
+    src_pitch = content.pitch
+    dst_pitch = ImageSurface.format_stride_for_width(cairo_format, content.width)
+    if not copy and dst_pitch == src_pitch and content.buffer != None :
+        pixels = content.buffer
+    else :
+        pixels = to_array(content, content.pixel_mode, dst_pitch)
+    result =  ImageSurface.create_for_data(
+        pixels, cairo_format,
+        content.width, content.rows,
+        dst_pitch)
+    return result
+
+def to_array(content, pixel_mode, dst_pitch = None) :
+    "returns a Python array object containing a copy of the Bitmap pixels."
+    if dst_pitch == None :
+        dst_pitch = content.pitch
+    buffer_size = content.rows * dst_pitch
+    buffer = array("B", b"0" * buffer_size)
+    dstaddr = buffer.buffer_info()[0]
+    srcaddr = cast(content.buffer, FT_Pointer).value
+    src_pitch = content.pitch
+    if dst_pitch == src_pitch :
+        memmove(dstaddr, srcaddr, buffer_size)
+    else :
+        # have to copy a row at a time
+        if src_pitch < 0 or dst_pitch < 0 :
+            raise NotImplementedError("can't cope with negative bitmap pitch")
+        assert dst_pitch > src_pitch
+        for i in range(content.rows) :
+            memmove(dstaddr, srcaddr, src_pitch)
+            dstaddr += dst_pitch
+            srcaddr += src_pitch
+    # pillow/PIL itself requires libtiff so it is assumed to be around.
+    # swap the bit-order from freetype's (MSB) to cairo's (host order) if needed
+    if ( ( byteorder == 'little' ) and (pixel_mode == FT_PIXEL_MODE_MONO ) ):
+        libtiff = CDLL("libtiff.so.5")
+        libtiff.TIFFReverseBits.restype = None
+        libtiff.TIFFReverseBits.argtypes = (c_void_p, c_int)
+        libtiff.TIFFReverseBits(buffer.buffer_info()[0], buffer_size)
+    return buffer
--- a/examples/bitmap_to_surface.py
+++ b/examples/bitmap_to_surface.py
@@ -1,13 +1,30 @@
+# FT_Bitmap to CAIRO_SURFACE_TYPE_IMAGE module
+# ============================================
+#
+# Copyright 2017 Hin-Tak Leung
+#
+# FreeType is under FTL (BSD license with an advertising clause) or GPLv2+.
+# Cairo is under LGPLv2 or MPLv1.1.
+#
+# This is a heavily modified copy of a few routines from Lawrence D'Oliveiro[1],
+# adjusted for freetype-py, and bugfix/workaround for mono-rendering [2].
+#
+# The bugfix/workaround requires libtiff on small-endian platforms.
+#
+# TODO: Look into using FreeType's FT_Bitmap_Convert() instead. However, libtiff
+#      is common enough, and probably not important.
+#
+#[1] https://github.com/ldo/python_freetype
+#    https://github.com/ldo/python_freetype_examples
+#
+#[2] https://github.com/ldo/python_freetype/issues/1
+#    https://github.com/ldo/python_freetype_examples/issues/1
+#
 '''
 FT_Bitmap to CAIRO_SURFACE_TYPE_IMAGE module
 ============================================
 
 Converting from Freetype's FT_Bitmap to Cairo's CAIRO_SURFACE_TYPE_IMAGE
-
-Copyright 2017 Hin-Tak Leung
-
-FreeType is under FTL (BSD license with an advertising clause) or GPLv2+.
-Cairo is under LGPLv2 or MPLv1.1.
 
 Usage:
      from bitmap_to_surface import make_image_surface
@@ -15,20 +32,6 @@ Usage:
 Works with cairocffi too. (Replace "from cairo ..." with "from cairocffi ...")
 
 Limitation: Surface.create_for_data is not in the "python 3, pycairo < 1.11" combo.
-
-This is a heavily modified copy of a few routines from Lawrence D'Oliveiro[1],
-adjusted for freetype-py, and bugfix/workaround for mono-rendering [2].
-
-The bugfix/workaround requires libtiff on small-endian platforms.
-
-TODO: Look into using FreeType's FT_Bitmap_Convert() instead. However, libtiff
-      is common enough, and probably not important.
-
-[1] https://github.com/ldo/python_freetype
-    https://github.com/ldo/python_freetype_examples
-
-[2] https://github.com/ldo/python_freetype/issues/1
-    https://github.com/ldo/python_freetype_examples/issues/1
 '''
 from freetype import FT_PIXEL_MODE_MONO, FT_PIXEL_MODE_GRAY, FT_Pointer, FT_Bitmap
 
--- a/examples/emoji-color-cairo.py
+++ b/examples/emoji-color-cairo.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+#  pycairo/cairocffi-based emoji-color example - Copyright 2017 Hin-Tak Leung
+#  Distributed under the terms of the new BSD license.
+#
+#  This script demonstrates overlapping emojis.
+#
+
+import freetype
+import numpy as np
+from PIL import Image
+
+from cairo import ImageSurface, FORMAT_ARGB32, Context
+
+face = freetype.Face("Apple Color Emoji.ttf")
+face.set_char_size( 160*64 )
+face.load_char('😀', freetype.FT_LOAD_COLOR)
+
+bitmap = face.glyph.bitmap
+width = face.glyph.bitmap.width
+rows = face.glyph.bitmap.rows
+bitmap = np.array(bitmap.buffer, dtype=np.uint8).reshape((bitmap.rows,bitmap.width,4))
+
+I = ImageSurface(FORMAT_ARGB32, width, rows)
+ndI = np.ndarray(shape=(rows,width), buffer=I.get_data(),
+                 dtype=np.uint32, order='C',
+                 strides=[I.get_stride(), 4])
+
+# Although both are 32-bit, cairo is host-order while
+# freetype is small endian.
+ndI[:,:] = bitmap[:,:,3] * 16777216 + bitmap[:,:,2] * 65536 + bitmap[:,:,1] * 256 + bitmap[:,:,0]
+I.mark_dirty()
+
+surface = ImageSurface(FORMAT_ARGB32, 2*width, rows)
+ctx = Context(surface)
+
+ctx.set_source_surface(I, 0, 0)
+ctx.paint()
+
+ctx.set_source_surface(I, width/2, 0)
+ctx.paint()
+
+ctx.set_source_surface(I, width , 0)
+ctx.paint()
+
+surface.write_to_png("emoji-color-cairo.png")
+Image.open("emoji-color-cairo.png").show()
--- a/examples/emoji-color-cairo.py
+++ b/examples/emoji-color-cairo.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
 #  pycairo/cairocffi-based emoji-color example - Copyright 2017 Hin-Tak Leung
--- a/examples/emoji-color-cairo.py
+++ b/examples/emoji-color-cairo.py
@@ -31,9 +31,12 @@ if ( face.glyph.bitmap.pitch != width * 4 ):
 bitmap = np.array(bitmap.buffer, dtype=np.uint8).reshape((bitmap.rows,bitmap.width,4))
 
 I = ImageSurface(FORMAT_ARGB32, width, rows)
-ndI = np.ndarray(shape=(rows,width), buffer=I.get_data(),
-                 dtype=np.uint32, order='C',
-                 strides=[I.get_stride(), 4])
+try:
+    ndI = np.ndarray(shape=(rows,width), buffer=I.get_data(),
+                     dtype=np.uint32, order='C',
+                     strides=[I.get_stride(), 4])
+except NotImplementedError:
+    raise SystemExit("For python 3.x, you need pycairo >= 1.11+ (from https://github.com/pygobject/pycairo)")
 
 # Although both are 32-bit, cairo is host-order while
 # freetype is small endian.
--- a/examples/emoji-color-cairo.py
+++ b/examples/emoji-color-cairo.py
@@ -8,6 +8,8 @@
 #
 #  Note: On Mac OS X before Sierra (10.12), change ttc->ttf;
 #        try Google's NotoColorEmoji.ttf at size 109 on Linux.
+#
+#  Limitation: Suface.get_data() is not in the "python 3, pycairo < 1.11" combo.
 
 import freetype
 import numpy as np
--- a/examples/emoji-color-cairo.py
+++ b/examples/emoji-color-cairo.py
@@ -6,6 +6,8 @@
 #
 #  This script demonstrates overlapping emojis.
 #
+#  Note: On Mac OS X before Sierra (10.12), change ttc->ttf;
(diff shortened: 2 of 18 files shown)


HinTak:
The liftiff-based workaround is not as bad as it sounds since python's PIL requires it. That Image.show() works at all means libtiff is around. Writing with FT_Bitmap_Convert likely will be slower.

HinTak:
Apparently Ubuntu/Debian has been shipping the region_t addition for some years for python 3 for the overlap check, so if you are on Ubuntu/Debian and using python 3, the wordle example may work as is without upgrading pycairo. Fedora certainly does not ship the patch.

HinTak:
wordle-cairo-fixed

under pycairo 1.11

HinTak:
wordle-cairo

with older pycairo.

HinTak:
wordle-cairo-pre

alpha drawing with older pycairo

HinTak:
wordle-cairo-post

alpha drawing with current pycairo.

HinTak:
This fixes the last issue; I don't particularly want to workaround the libtiff dependency for mono(1-bit) drawing, so this is all.

HinTak:
diff

To try to get an exact match, I changed the number of words back to 100, seed the random number to be the same, plus this change (the original example flp the y axis a few times...):

```
--- a/examples/wordle-cairo.py
+++ b/examples/wordle-cairo.py
@@ -153,7 +153,7 @@ if __name__ == '__main__':
             for dx,dy in spiral():
                 c = .25+.75*random.random()
                 x = int(x0+dx)
-                y = int(y0+dy)
+                y = H - int(y0+dy)
                 checked = False
                 I.flush()
                 if not (x <= w//2 or y <= h//2 or x >= (W-w//2) or y >= (H-h//2)):
```

and get this comparison from imagemagick's compare.

This shows that the first few tiles from the center of the spiral are identical, but pixel differences do accumulate, and after about 5 tiles, the differences are enough that the tiling of the spirals no longer matches. Just for interest.

HinTak:
glyph-lcd-cairo

output of lcd example. I added the lcd_v case also.

HinTak:
glyph-vector-cairo-1

The output of the re-written glyh-vector example. I'll tidy up the code and push at some point.
However I found the outlines is strictly speaking wrong (https://github.com/rougier/freetype-py/issues/58)
and the relative scale of glyph-vector-2 between the outline and the bitmap is also wrong 
( https://github.com/rougier/freetype-py/issues/59 ) .

HinTak:
glyph-vector-2-cairo

The bitmap + outline example; currently still in png; need to add some svg related code...

HinTak:
Finished. Please review and merge. @rougier

HinTak:
emoji-color-cairo

sorry, one more example. here is the outcome. quite cool!

Definitely stopping now.

freetype/__init__.py
```diff
@@ -650,8 +650,7 @@ def get_cbox(self):
         which is dedicated to this single task.
         '''
         bbox = FT_BBox()
-        error = FT_Outline_Get_CBox(byref(self._FT_Outline), byref(bbox))
-        if error: raise FT_Exception(error)
+        FT_Outline_Get_CBox(byref(self._FT_Outline), byref(bbox))
```
rougier: Why did your remove the error message?

examples/glyph-vector.py
```diff
@@ -5,6 +5,12 @@
 #  FreeType high-level python API - Copyright 2011 Nicolas P. Rougier
 #  Distributed under the terms of the new BSD license.
 #
+#  - The code is strickly speaking not optimal, as it ignores the 3rd order
```
rougier: I think this could be rephrased

examples/glyph-vector-2.py
```diff
@@ -5,6 +5,12 @@
 #  FreeType high-level python API - Copyright 2011-2015 Nicolas P. Rougier
 #  Distributed under the terms of the new BSD license.
 #
+#  - The code is strickly speaking not optimal, as it ignores the 3rd order
```
rougier: This could be rephrased

examples/emoji-color-cairo.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
```
rougier: python3 or python ?
Why do you add a shebang for every script?

HinTak: That came from the old one (the old emoji-color). I 'll change it to plain python now. I like to shebang if it can run as a script (not all python code does), rather than intended for import.

freetype/__init__.py
HinTak: As in the commit message! FT_Outline_Get_CBox does not return a value! It segfault as you wrote it!

examples/glyph-vector.py
HinTak: What do you suggest?

examples/glyph-vector-2.py
HinTak: What do you suggest?

HinTak:
@rougier : updated the shebang and answered your question on cbox. Please feel free to update and rephrase the comment on cubic curve after. Please merge.

rougier:
I've started the review and I also need to test all the new examples. But in the meantime, thank you, very nice work.

I think it might be best to split the cairo / matplotlib example into two directory (cairo and matplotlib) and to keep the same name for the example. At the root, we can leave only the examples that does not depends neither on cairo nor on matplotlib.

HinTak:
Most of the samples use Vera.ttf in the current directory, so need adjusting if moved. Yes, the -cairo gets a bit tedious. -c?

rougier:
I was thinking more of `cairo/glyph-vector.py` and `matplotlib/glyph-vector.py`

examples/emoji-color-cairo.py
```diff
@@ -0,0 +1,54 @@
...
+#  This script demonstrates overlapping emojis.
+#
+
+import freetype
+import numpy as np
+from PIL import Image
+
+from cairo import ImageSurface, FORMAT_ARGB32, Context
+
+face = freetype.Face("Apple Color Emoji.ttf")
```
HinTak: Yes, the other emoji example hard coding apple's path. I tried google noto but less pretty. See later commit comment

examples/glyph-vector-2-cairo.py
```diff
@@ -0,0 +1,152 @@
...
+        elif (CODES[i] == LINETO):
+            ctx.line_to(VERTS[i][0],VERTS[i][1])
+            i += 1
+        elif (CODES[i] == CURVE3):
+            ctx.curve_to(VERTS[i][0],VERTS[i][1],
+                         VERTS[i+1][0],VERTS[i+1][1], # undocumented
+                         VERTS[i+1][0],VERTS[i+1][1])
+            i += 2
+        else:
+            raise NotImplementedError("Cannot cope!")
```
HinTak: Placeholder for when I add cubic curves :-)...

examples/emoji-color-cairo.py
```diff
@@ -0,0 +1,54 @@
...
+bitmap = np.array(bitmap.buffer, dtype=np.uint8).reshape((bitmap.rows,bitmap.width,4))
+
+I = ImageSurface(FORMAT_ARGB32, width, rows)
+ndI = np.ndarray(shape=(rows,width), buffer=I.get_data(),
+                 dtype=np.uint32, order='C',
+                 strides=[I.get_stride(), 4])
+
+# Although both are 32-bit, cairo is host-order while
+# freetype is small endian.
+ndI[:,:] = bitmap[:,:,3] * 16777216 + bitmap[:,:,2] * 65536 + bitmap[:,:,1] * 256 + bitmap[:,:,0]
```
HinTak: No, your suggested change seems wrong - does it actually work? Mine was an implied double-loop. It loops over the two indices on the C side of the interpreter so it is a lot faster than an explicit double-loop.

examples/emoji-color-cairo.py
HinTak: I could use the full apple path as the other emoji example does? google noto emoji font has a bigger problem - its location differs between linux distros. So I'd usethe full apple authoritative path and hope for the best, really.

examples/bitmap_to_surface.py
```diff
@@ -0,0 +1,95 @@
+'''
```
HinTak: okay

examples/example_1-cairo.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # load glyph image into the slot (erase previous one)
+        charcode = ord(text[n])
+        index = FT_Get_Char_Index( face, charcode )
+        FT_Load_Glyph( face, index, FT_LOAD_RENDER )
+
+        # now, draw to our target surface (convert position)
+        draw_bitmap( slot.contents.bitmap,
+                     slot.contents.bitmap_left,
+                     HEIGHT - slot.contents.bitmap_top )
```
HinTak: why -1? it is an offset and starts from zero.

examples/glyph-lcd-cairo.py
```diff
@@ -0,0 +1,115 @@
...
+
+    face = Face('./Vera.ttf')
+    face.set_char_size( 48*64 )
+    face.load_char('S', FT_LOAD_RENDER |
+                        FT_LOAD_TARGET_LCD )
+    bitmap = face.glyph.bitmap
+    width  = face.glyph.bitmap.width//3
+    rows   = face.glyph.bitmap.rows
+    pitch  = face.glyph.bitmap.pitch
+    whole_thing = (c_ubyte * (pitch * rows))()
```
HinTak: okay.

examples/glyph-lcd-cairo.py
```diff
@@ -0,0 +1,115 @@
...
+                  offset=1,
+                  strides=[pitch, 3])
+    ndB = ndarray(shape=(rows,width), buffer=whole_thing,
+                  dtype=ubyte, order='C',
+                  offset=2,
+                  strides=[pitch, 3])
+    ndI = ndarray(shape=(rows,width), buffer=I.get_data(),
+                  dtype=uint32, order='C',
+                  strides=[I.get_stride(), 4])
+    ndI[:,:] = 4278190080 + ndR[:,:] * 65536 + ndG[:,:] * 256 + ndB[:,:]
```
HinTak: I'd add a comment next to it saying what those numbers are, best of both worlds?

examples/glyph-lcd-cairo.py
```diff
@@ -0,0 +1,115 @@
...
+    rows_old = rows
+
+    # LCD_V
+    face.load_char('S', FT_LOAD_RENDER |
+                        FT_LOAD_TARGET_LCD_V )
+    bitmap = face.glyph.bitmap
+    width  = face.glyph.bitmap.width
+    rows   = face.glyph.bitmap.rows//3
+    pitch  = face.glyph.bitmap.pitch
+    whole_thing = (c_ubyte * (pitch * face.glyph.bitmap.rows))()
```
HinTak: okay, the name was daft. I'll change.

examples/glyph-vector-2-cairo.py
HinTak: I have some new code near that location to get 3rd order curves (what makes the latest figures at
https://github.com/rougier/freetype-py/issues/58 )

examples/glyph-vector.py
HinTak: 'not optimal' isn't quite the right word - on rethinking, I think 'incomplete' might be better regarding the interpolation and assuming first point is on. The spltting 3rd-order curve into two 2nd order one is strictly speaking a bug, but can only be seen if you look at a postscript font (such as those shipped with adobe acrobat reader - see latest figures on https://github.com/rougier/freetype-py/issues/58 ).

examples/glyph-vector-2.py
HinTak: It currently misprocess 3rd order curves.

HinTak:
Yes, moving into drectories could be an option especially now it is getting quite a few.

examples/glyph-lcd-cairo.py
```diff
@@ -0,0 +1,115 @@
...
+                  offset=pitch,
+                  strides=[pitch*3, 1])
+    ndB = ndarray(shape=(rows,width), buffer=whole_thing,
+                  dtype=ubyte, order='C',
+                  offset=2*pitch,
+                  strides=[pitch*3, 1])
+    ndI = ndarray(shape=(rows,width), buffer=I.get_data(),
+                  dtype=uint32, order='C',
+                  strides=[I.get_stride(), 4])
+    ndI[:,:] = 4278190080 + ndR[:,:] * 65536 + ndG[:,:] * 256 + ndB[:,:]
```
HinTak: changed my mind - python should cache constant values, so there should be no difference with `256**3`. I'll time it and see.

examples/emoji-color-cairo.py
rougier:
```
Z = np.random.uniform(0,1,(3,3,4))
R1 = Z[:,:,3]*4 + Z[:,:,2]*3 + Z[:,:,1]*2 + Z[:,:,0]
R2 = (Z*(4,3,2,1)).sum(axis=-1)
print(np.allclose(R1,R2)
```

examples/example_1-cairo.py
rougier: if your 'y' value goes from `0` to `HEIGHT-1` then `HEIGHT-y` goes from `1` to `HEIGHT`.

examples/emoji-color-cairo.py
rougier: But your version is faster actually

HinTak: Please do not make suggestion without testing. Not writing as I wrote tends to recreate a new numpy array and breaks the association between the cairo surface with the calculation output. i.e. you finish your calculation but the result does not get written to the cairo surface. Please test before you make suggestions.

HinTak: Speed is one thing, but I did not write it that way for it to be fast. The calculation needs to work end up getting written in the right place.

rougier: The`Z` example was only to show you the result is the same. My suggestion does write to the underlying memory buffer (because in your case `... = :,:`)

rougier:
A second PR to moving things into directories would be best, let's not overloading this one.

examples/emoji-color-cairo.py
HinTak: No, the calculation result might be the same, but the outcome is **not** the same. You don't understand that line of code. Please stop making suggestion about changing that line wtithout testing that your suggestion works as a substitute. It was hours of work.

examples/example_1-cairo.py
HinTak: ? y does not go from 0 to HEIGHT -1 - it is a dicrete set of values, with zero being minimum.

HinTak:
I wasn't going to - I don't mind it either way, and wouldn't mind if you want to move; but I am not going to move within this (nor going to make another separate PR for move).

--------------------------------------------
On Wed, 3/5/17, Nicolas P. Rougier <notifications@github.com> wrote:

 A second PR to moving things into
 directories would be best, let's not overloading this
 one.

examples/emoji-color-cairo.py
rougier: What reshape are you talking about exactly ? The suggestion uses numpy broadcasting  and a simple sum over the last axis then copy the result inplace in the array, taking order and strides into account. Anyway, no need to answer, we can keep the line as it is currently.

However, I cannot test it (yet) because `ImageSurface.get_data()` has been introduced in cairo 1.2. You should check for version at the start of each example to tell user he needs to upgrade (else it is difficult to identify the problem).

examples/example_1-cairo.py
rougier: I meant: if `slot.contents.bitmap_top` possible values are in the range `[0, HEIGHT-1]`, then `HEIGHT - slot.contents.bitmap_top` possible values are in the range `[1,HEIGHT]`. If the `draw_bitmap` function takes the y argument as being valid between `1` and `HEIGHT` then fine.

examples/glyph-vector.py
rougier: I agree, we're not talking about optimality. "incomplete" or "incorrect for some rare cases" would be better.

examples/emoji-color-cairo.py
rougier: Silly question, I can only find "Apple Color Emoji.ttc" on my system (Sierra 10.12.4), is there any difference ?

examples/example_1-cairo.py
HinTak: No, HEIGHT and slot.contents.bitmap_top are unrelated. the minium value of slot.contents.bitmap_top is zero I think, the max is probablt glyph.bbox.xMax or something of that sort. HEIGHT is the size of the whole drawing, and arbitrary - defined at the top of the file.

Also cairo can draw shapes partly off canvas - I don't think the code would fail even if you pass a large negative value: what it being draw is just partly or whole off-canvas.

HinTak:
@rougier  Added 8 commits, a few to do with the review and a few fixes the cubic curve/alignments. I don't think there are any important issues; so I'd say just merge.

rougier:
Nope, still need test and to test, I need to instal pycairo > 1.10. Unfortunately the "official" version from https://www.cairographics.org/pycairo/ seems to be stuck at 1.10.0. What version do you use actually?

HinTak:
Although I mentioned at the beginning pycairo 1.11+ is needed, I have since worked around that.
If you are on python 2.x, python 1.10 is suficient - I think I ran all of them with just 1.10 as shipped by Fedora
just fine, with python 2.x . 

pycairo 1.10 for python 3.x lacks one or two things so, If you are on python 3.x,
a few examples (the later/more complicated ones)
needs python 1.11+ from https://github.com/pygobject/pycairo .

--------------------------------------------
On Thu, 4/5/17, Nicolas P. Rougier <notifications@github.com> wrote:

 
 Nope, still need test and to
 test, I need to instal pycairo > 1.10. Unfortunately the
 "official" version from https://www.cairographics.org/pycairo/
 seems to be stuck at 1.10.0. What version do you use
 actually?

examples/emoji-color-cairo.py
HinTak: ttc is larger and has a 2nd typeface (.e. `Face("Apple Color Emoji.ttc", 1)`). Anyway, I have added another commit t to use the full sierra path, asI think it should either work on a current and correct os, or not at all, and added a comment about older Mac Os and Linux.

examples/glyph-vector.py
HinTak: I have replaced "strickly speaking not optimal" in a latter commit to "incomplete and over-simplified".

examples/emoji-color-cairo.py
HinTak: No, `ImageSurface.get_data()` is lackin only in the "python 3 + pycairo < 1.11" combo. Already noted in the comment. python 2 with antique version of pycairo is fine. The limitation is noted in the comment (in a later commit).

I suppose I could add a test for the presence of  `ImageSurface.get_data()`. If I can get it to work, I'll just push another commit . The comment about limitation should be sufficient.

HinTak: Okay, I have added a `try catch` block to the 3 examples which needs `get_data` around the first use.
https://github.com/rougier/freetype-py/pull/55/commits/2f8f1a9a336a93b6dd5702935093d81c41306060

it now says:
```
For python 3.x, you need pycairo >= 1.11+ (from https://github.com/pygobject/pycairo)
```
It is a limitation of the pythn 3 + older pycairo combo. python 2 + antique pycairo is fine.

Two of the 3 examples already said so at the comments at the top. So I added a comment to the 3rd:

https://github.com/rougier/freetype-py/pull/55/commits/81c32958f8434f20c127fb0b211f379e460fdfbb

HinTak:
Beside the `try catch`, I had a look at what python 3's older pycairo is exactly doing, and the change log.
That "Surface.get_data: Not Implemented yet." itself is a typo. Somebody just forgot to change
"Surface.get_data" to "ImageSurface.get_data" . This method is present but always throws in the
in the python 3 + older pycairo combo. I wish it is just missing. As it is, one  only know by trying to run that method and to see if it works.
Not pretty.

Version check is unreliable - python 2 + pycairo 1.10.0 works, python 3 + pycairo 1.10.0 (stock) does not. Besides, some distros carry extra patches for 1.10.0 ... so even the python 3 + pycairo 1.10.0 combo may work, depending on what distro you use. So there isn't much one can do, beyond just try and catch the actual run. At least now one get told pycairo is too old, and where to download.

rougier:
No luck in installing pycairo on OSX. Need to investigate. I'll merge and see that later.

HinTak:
That's a pity. Do you already have cairo instead from homebrew or macport? It is not trivial to build cairo by hand. The last time I cross-compile cairo for mac was a bit over a year ago. You need to have libpng, freetype, fontconfig, pixman, first, in that order. and possibly some of g*, such as glib2 . After that then you can talk about pycairo. I suspect you can get pycairo via homebrew/macport? That's easier. May not be the latest but as I said, if you use python 2.x, the antique version pycairo works. Seems max os x still ships multiple versions of python? So many things on linux depend on cairo, it is quite trivial on linux.

rougier:
I'm using Python 3. I tried the homebrew version (that works) but it is stuck at 1.10. Same from pip.

From latest release, I got an error when building:

```
duplicate symbol _Pycairo_Content_Type in:
    build/temp.macosx-10.12-x86_64-3.6/cairo/error.o
    build/temp.macosx-10.12-x86_64-3.6/cairo/cairomodule.o
```

I need to investigate when I've some free time.

rougier:
It was a missing extern in the end. Build is fine now.

rougier:
Shadow is missing in the `emoji-color-cairo.py` in the final png output. But the  `Image.open(...).show()` tends to suggest it is here (even if display is wrong)

`Image.open(...).show()`:

untitled

`emoji-color-cairo.png`:
emoji-color-cairo

Note there are spurious pixels below smileys (you have to zoom to see them)

HinTak:
I see you already filled the sierra build issue and had it merged ( https://github.com/pygobject/pycairo/pull/36 ). I was going to filed & @ you to supply the details. Turn out I haven't pulled since clone so my pycairo is still at v1.12.0-5-g29805af at two weeks ago; the line affecting you was added on Tuesday this week. Wouldn't have helped as that still builds fine on linux.

You have found an interesting issue with the emoji font. version 10.0d3e2 (shipped with Yosemite I think, the one I usually test with) is okay; I grabbed a ttc from the internet, supposedly 12.0d17e1 , and it shows as yours - `Image.open(...).show()` shows the extra shadows below, and indeed there are a few stray pixels.

The stray pixels are real; but the  `Image.open(...).show()` extra-shadow appearance seems to be a bug with  `Image.open(...).show()`.  On Linux,  `Image.open(...).show()` seems to be calling pngtopam and convert to a pam in the /tmp directory, them run imagemagick's `display` on the temp file. I can get the extra shadow by running the pngtopam on the 12.x version png mannually. I think it is a bug with pngtopam. gimp shows the transparent background, the stray pixels, but not the extra-shadow; flattening the image does not make the shadow appear.

rougier:
Using ImageMagick and converting the png to ppm I get the broken image so it's unlikely a bug if the same extra shadow appears using two different converters.

HinTak:
It is not two different converters - imagemagick calls pngto* . Try moving your pngto* programs aside, and you will find that imagemagick stops working. You are seeing the same bug. It is one bug, in  pngtopam . Try using somerthing that does not call pngto* .

rougier:
Didn't know that. Using the `-flatten` flag seems to fix the bug.

HinTak:
I used xv and gimp to looked at it also. xv is older than imagemagick so I am fairly sure it doesn't use it, but I don't expect you to have xv since it is X11 based. gimp doesn't show the extra shadow either. Surely on Mac OS X there are image viewers which does not use imagemagick?

Since python PIL's uses ` Image.open("emoji-color-cairo.png").show()` calls imagemagick's `display`, I am fairly sure they are just the same thing. At what point imagemagick get confused I don't know.

HinTak:
It seems that I was wrong. PIL has its own png reader ( PIL/PngImagePlugin.py ) . So we do have two sets of applications, PIL and imagemagick on oneside, xv and gimp on the other, which reads the png differently.

`Image.show` 's docstring mentions that it uses `display` to to display on unices.

```
        On Unix platforms, this method saves the image to a temporary
        PPM file, and calls either the **xv** utility or the **display**
        utility, depending on which one can be found.
 
        On macOS, this method saves the image to a temporary BMP file, and opens
        it with the native Preview application.

```

HinTak:
Okay, I think I got to the problmem with the alpha channel. PIL Image has its own image writing code ( PIL/PpmImagePlugin.py , PIL/BmpImagePlugin.py ) to saves to bmp or ppm , before launching the display code. The image writing code ignore alphas. So the thing is that programs which ignores the alpha channel sees the extra shadow. And you do have 3 programs which ignore the alpha channel, convert in its default mode, PIL, and netpbm-progs. 

Apple put some garbage there and also says that part is transparent, that's pretty much it.

HinTak:
Issuing a pull for the display problem - @rougier : would be nice if you get it a try on Mac os .
https://github.com/python-pillow/Pillow/pull/2509

PR #60, opened 2017-06-09
Allow Type 14 charmap to fail in FT_Set_Charmap.

HinTak:
FT2 demo's ftdump ignores the return code of FT_Set_Charmap.

FT_Set_Charmap is known to fail when type~14 charmap is selected
(which doesn't map character codes to glyph indices at all).
This is a harmless situation. OpenType 1.8 fonts with Variations shows:

"format 14, platform 0, encoding  5   (Unicode Variation Sequences)"

An example is the Google Noto CJK /Adobe Source Sans fonts.

Also minor addition to API: auto-detect input type, and allow input as
integer index also.

### DIFF
--- a/freetype/__init__.py
+++ b/freetype/__init__.py
@@ -1097,9 +1097,16 @@ class Face( object ):
         '''
         Select a given charmap for character code to glyph index mapping.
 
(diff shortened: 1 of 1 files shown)


rougier:
Thanks

PR #64, opened 2017-09-17
sync Error code list to FreeType 2.8.1

HinTak:
Not sure why your 0xB3 onwards are different, but these are sync to current.


freetype/ft_errors.py
```diff
@@ -90,15 +93,22 @@ def __str__(self):
         0x99: "could not find context" ,
         0x9A: "invalid PostScript (post, table format" ,
         0x9B: "invalid PostScript (post, table" ,
+        0x9C: "found FDEF or IDEF opcode in glyf bytecode" ,
```
rougier: glyph

HinTak: "glyf" is the name of a opentype table. The above is correct. See:

https://www.microsoft.com/typography/otspec/glyf.htm

rougier: ok

Issue #65, opened 2017-11-17
trim symbol list down to match visual studio window build of freetype


HinTak:
See
https://github.com/rougier/freetype-py/issues/17#issuecomment-345205940

It looks like the typical visual studio build of freetype does not include support for some of the more unix-centric font formats. Consider trimming the symbol list down, although this is arguable since the typical freetype-py users are unix-based.

rougier:
Is there a liste of such symbols somewhere ?

HinTak:
Freetype ships a visual studio project file (contributed from somebody else. )It should be possible to do a visual studio build from it and read the symbol list. I can do it at some point.

rougier:
Thanks for the offer. Is it worth the effort?

HinTak:
It is not much an effort. I ships mingw cross-compiled Freetype because it is easier for me and it matches the Unix feature set better, but the visual studio route needs to be looked at, at some point, since that's more natural to windows developers. So I'll do it at some point and perhaps push some suggestions upstream also if I finish anything significantlo missing in vs-built Freetype.

rougier:
Ok, thanks.

HinTak:
It is actually not too bad - my mingw-cross compiled freetype have only 5 more symbols, these BDF ones and

```
FT_Get_CID_From_Glyph_Index
FT_Get_CID_Is_Internally_CID_Keyed
FT_Get_CID_Registry_Ordering_Supplement
``` 

I'll go about asking the freetype people adding those 5 to keep them in syn, but for how just protecting the *BDF* nes should do since the CID ones are not in freetype-py yet.

PR #66, opened 2017-11-27
Protect symbols not in default visual studio build of freetype 2.8.x

HinTak:
See
https://github.com/rougier/freetype-py/issues/65


rougier:
Thanks.

PR #67, opened 2018-02-01
Add information to readme for FreeType installation on windows

drunsinn:
Extended the Windows-section of the readme file. Some of the information is also availible on the issue page.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -28,10 +28,36 @@ the library binary file in '/usr/local/lib'.
 Window users
 ------------
 
-You can try to install a window binaries available from the Freetype site or
-you can compile it from sources. In such a case, make sure the resulting
-library binaries is named 'Freetype.dll' (and not something like
-Freetype245.dll) and make sure to place a copy in Windows/System32 directory.
+There are no official Freetype binary releases available, but they offer some
+links to precompiled Windows DLLs. Please see the `FreeType Downloads
+<https://www.freetype.org/download.html>`_ page for links.
+You can also compile the FreeType library from source.
+
+32-Bit vs 64-Bit on Windows
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you are using freetype-py on Windows with a 32-Bit version of python, you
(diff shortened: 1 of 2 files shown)


rougier:
Thanks @drunsinn.  Can you take @HinTak comment into account (freetype does not offer binary?

drunsinn:
I changed the wording to be more specific about the fact that FreeType dose not offer a binary release for windows.
In addition to that I went with @HinTak 's recommendation and explained more about where to place the dll. At first i also wanted to mention regsvr32, but that dose not work with any FreeType dll i could find.

Another way I didn't mention in the text was to edit the environment variable inside the python application (os.environ["PATH"] += os.pathsep + path)
Has anyone of you had experience with this?

HinTak:
The issue with editing environment variable within an application is generic - it is an "environment" variable, and as such, only have effect to child processes spawned by the current one, and has no effect on the current one. I think it take special measures to force the current process's PATH variable to change. It is possible on unix systems, I don't know the exact api off the top of my head, but I know where to look it up. I don't know the details for windows.

regsvr32 is irrelevant - freetype's dll does not work that way as a "com server", in windows-lingo.

drunsinn:
I don't see the temporary effect of changing os.environ["PATH"] as a negative thing. This way ist is easy for an application to make sure the correct library (32bit vs 64bit in platform.architecture()[0]) is loaded.

On the other hand, you are completely right, the ‘runtime’ solution shouldn't be mentioned in the documentation. Things like the manipulation of an environment variable should be done only once during the installation process and not every time the application is run.

HinTak:
I wasn't saying manipulating os.environ["PATH"] within a program is a negative thing - I was just trying to outline what precisely the current behavior is: changes to PATH affects child processes spawned by the current process, not the current process. There are special calls to make PATH take effect in the current process, but it is a special extra step.

rougier:
Thanks @drunsinn for the PR, thanks @HinTak for the review.

PR #69, opened 2018-02-06
Avoid segfault on passing non-binary data to FT_Get_Name_Index

madig:
I haven't looked at the source closely, but more functions may need a vetting.


rougier:
Thanks !

PR #73, opened 2018-02-23
Fix #72: install on Python 3.6 on Windows

### LINKED ISSUE #72: Cannot install 1.2 on Python 3.6, Windows 10 x86_64
```
pip install --user -U freetype-py
Collecting freetype-py
  Downloading freetype-py-1.2.tar.gz (181kB)
    100% |████████████████████████████████| 184kB 6.2MB/s
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "C:\Users\NIKOLA~1.WAX\AppData\Local\Temp\pip-build-c3iwzy2b\freetype-py\setup.py", line 12, in <module>
        'README.rst')).read()
      File "c:\program files\python36\lib\encodings\cp1252.py", line 23, in decode
        return codecs.charmap_decode(input,self.errors,decoding_table)[0]
    UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 3934: character maps to <undefined>
```

belluzj:
Force the encoding = "utf-8" instead of the default Windows codec

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ from codecs import open
 from setuptools import setup, Extension
 
 description = open(path.join(path.abspath(path.dirname(__file__)),
-                             'README.rst')).read()
+                             'README.rst'), encoding='utf-8').read()
 
 setup( name        = 'freetype-py',
        version     = '1.2',


madig:
It would be nice to have a CI service test every commit with tox for py27 and 35 or something, maybe with just a simple test to see if we can load a FT_Face. Can have a look when I get the time...

rougier:
@belluzj Thanks for the fix.
@madig Yes, a CI service would be great.

PR #76, opened 2018-02-24
README.rst: fix invalid markup that fails to render on PyPI

### LINKED ISSUE #75: README.md instead of README.md
It seems [Pypi](https://pypi.org/project/freetype-py/) does not interpret restructured text but interpret markdown. Hence, it would be better to use a `README.md`. 

The list of contributors should be updated also.

anthrotype:
Fixes #75 

I checked by installing `readme_renderer` and running `python setup.py check --restructuredtext --strict`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -59,7 +59,9 @@ system folder, you can choose one of the following ways to solve this:
 
 To get a complete list of all the directories in the $PATH
 environment variable (user and system), open a command promt and type
+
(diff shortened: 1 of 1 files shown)


rougier:
Thanks. Not sure it's worth a release. What do you think?

anthrotype:
I don’t think so. It’ll be fixed upon the next one

PR #77, opened 2018-02-24
Setup CI testing

madig:
First stop: add basic smoke test for py27, py36. The test only sees if stuff loads after a tox-controlled install.

Next: Setting up some CI service to tox on Windows, macOS and some Linux. That's probably the hard part.

### DIFF
--- a/tests/test_ft_face.py
+++ b/tests/test_ft_face.py
@@ -1,0 +1,6 @@
+import freetype
+
+
+def test_ft_face():
+    """A smoke test."""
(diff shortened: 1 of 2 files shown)


rougier:
Nice, thanks. That's a good start.

madig:
Waaah, I wanted to push more stuff here :grin:

rougier:
Oops, sorry. But now Travis is on so we can check if it works with new PR.

PR #79, opened 2018-03-02
WIP: Setup CI

madig:
We want a bundled shared FreeType library with a pinned version. If none comes bundled with the package, look for the system-wide one.

I copied and adapted what's done in ttfautohint-py, more testing needs to be done on macOS and Windows. Currently, no harfbuzz is built-in, not sure how to handle it yet (circular dependency freetype <-> harfbuzz).

Once the wheel build stuff works everywhere, I can experiment with Travis and Appveyor.


freetype/raw.py
```diff
@@ -12,23 +12,35 @@
...
+else:
+    library_name = 'libfreetype.so.6'
 
-if filename is None:
-    if platform.system() == 'Windows':
-        # Check current working directory for dll as ctypes fails to do so
-        filename = os.path.join(os.path.realpath('.'), 'freetype.dll')
-    else:
-        filename = 'libfreetype.so.6'
+filename = pkg_resources.resource_filename('freetype', library_name)
```
anthrotype: using `pkg_resources` adds a runtime dependency on setuptools -- which nowadays is preinstalled on most python distributions alongside pip, but technically is not part of the python standard library.
The old-school approach to get a resource filename relative to `__file__` works fine, unless you want to get a filename of a data file from a inside a zipped python package (like an .egg). But since we are including a precompiled native library, we have to set `zip=False` in setup.py in any case, otherwise we can't load it. So using pkg_resources has no concrete advantage IMO over `os.path.dirname(__file__, library_name)`

anthrotype:
> no harfbuzz is built-in, not sure how to handle it yet (circular dependency freetype <-> harfbuzz)

I guess you;ll have to build freetype --without-harfbuzz, then build harfbuzz --without-freetype, and finally freetype --with-harfbuzz

anthrotype:
correction: of course, I meant harfbuzz --with-freetype

anthrotype:
then we also need to figure out a way to make this optional, as there are some users that would like to simply `pip install .` or `python setup.py install` without building the embedded library, but using the system installed one (the current default behavior).
One way could be to add a command-line option or read from an environment variable that tells the setup.py to not add any `ext_modules` thus making the package "pure" python.

anthrotype:
another thing with this setup borrowed from ttfautohint-py, is that the source distribution as generated with `python setupt.py sdist` will not include the freetype2 source. So, for example, running `tox` won't work, because tox by default creates the sdist, then installs it into the virtual environment, before running the tests inside it (ok, one can also do `tox --develop` or set `usedevelop=True` in `tox.ini`). 

In any case, building the embedded library currently requires access to the freetype2 git submodule (thus one has to `git clone --recursive ...` or do `git submodule update --init` afterwards). And this submodule is not included in the python package sdist. I don't think it would be a good idea trying to add all the freetype2 sources from the git submodule to the python sdist package, because freetype's own build system probably detects whether it's running from a git repository or a released tarball.

One option instead of using git submodules would be to have the setup.py itself download the freetype2 released tarball from the official website, and do the build from there.
Using the released tarballs also has the advantage that you won't need to do `./autogen.sh` but can do `./configure` directly, so you don't need all the autotools but just make.

Actually, the only reason I used git submodules for ttfautohint-py was that at the time I wrote it, the stable ttfautohint release didn't have an option to build a shared library. If I were to redo it now, I would rather use the stable releases instead of the git repositories.

madig:
Hm, running tox is actually on my want-list, so I guess I need to start looking into downloading a tarball...

Thanks for your input on this :)

Will send a PR next to move freetype/ into src/python/.

madig:
Alright, using a tarball works on Fedora 27. Now testing on macOS 10.13.

madig:
It now builds and install the module on Fedora 27 and macOS 10.13, running `lsof | grep freetype` confirms on both that the packaged library is used.

madig:
I just did the move from freetype/ to src/freetype and added the Makefile to the manifest, running `tox` now passes :)

.travis.yml
```diff
@@ -0,0 +1,68 @@
...
+        - MB_PYTHON_VERSION=3.6
+
+before_install:
+    - source ci/multibuild/common_utils.sh
+    - source ci/multibuild/travis_steps.sh
+    - before_install
+
+install:
+    # Maybe get and clean and patch source
+    - clean_code $REPO_DIR $BUILD_COMMIT
```
anthrotype: You don’t need this line

ci/multibuild_config.sh
```diff
@@ -0,0 +1,6 @@
+# Define custom utilities
+# Test for OSX with [ -n "$IS_OSX" ]
+
+function run_tests {
+    tox
```
anthrotype: This is going to build the shared library from source again, whereas you actually want to test the precompiled wheel (the one compiled inside the manylinux1 docker).
Use tox with —installpkg option, or just skip tox and run the tests directly. Multibuild already makes sure you are running the run_tests function from inside a fresh venv with the built wheel installed.
You can also lists tests requirements with an env variable

ci/multibuild_config.sh
```diff
@@ -2,5 +2,5 @@
 # Test for OSX with [ -n "$IS_OSX" ]
 
 function run_tests {
-    tox
+    pytest
```
anthrotype: Forgot to mention (though it’s in multibuild docs) that this function runs from within a new temporary subfolder, so if you want to run pytest from the root of th repo, you need to cd ..

anthrotype: https://github.com/fonttools/ttfautohint-py/blob/f099836e367bbc0d8f6652ac23f8e849d5b366b6/ci/multibuild_config.sh#L45

anthrotype:
It’s working! 👍

madig:
Okay, more work some other day. Get things to deploy on PyPI or get Appveyor up first?

anthrotype:
Appveyor first. PyPI deployment is more complicated

madig:
Kay :) 

Oh, well, still need to give the option to not bundle FT. Maybe look for an env var and bundle FT only then? The var can be set in the build files. The default would be not to bundle.

anthrotype:
@rougier needs to enable the repository with his Appveyor account for this to work, just like he did for Travis.

Yes, default should be not bundle

madig:
Is it possible to cache the FT tarball somewhere? I feel uncomfortable downloading 2 Mb off the internet multiple times for every commit :(

madig:
I'll be using https://ci.appveyor.com/project/madig/freetype-py until @rougier enables Appveyor on his repo :)

anthrotype:
> Is it possible to cache the FT tarball somewhere?

You could make a separate make target that just curls the tarball and make the other target depend on it

anthrotype:
Oh you mean for every commit.. hm yes Travis has a cache feature, check the docs

madig:
Will do.

The Windows MSYS2 build produces a libfreetype.dll.a, file says it is a "current ar" -- is this related to the dllwrap magic you did?

anthrotype:
That’s just the import library (equivalent of *.lib when building with MSVC) for use when linking, you don’t need that.
You don’t get any DLL? Try configuring with —enable-shared, maybe it’s not enabled by default on MSYS2 builds.

anthrotype:
I suggest you do your experiments on a real Windows machine or a local VM first, it’s easier.

Also use Dependency Walker to check that the built dll doesn’t depend on anything else but system libraries, we want to bundle a single freetype DLL.

http://www.dependencywalker.com

If you see it still depends on msys2 specific dlls, then try to create the DLL using the dllwrap tool and see if that makes it a standalone library.

(I’m at the airport typing from my phone)

anthrotype:
Maybe also try passing `-static-libgcc`
https://stackoverflow.com/a/18138926

anthrotype:
Also, technically we don’t _have_ to use kings-w64 gcc to build the freetype DLL on windows, we could also build it with MSVC (there should be an nmake makefile in win32 folder). The reason I had to use msys2 for ttfautohint-py was because ttfautohint depends on gnulib and that can only be built with the autotools in a Unix-like environment.

anthrotype:
mingw, not kings

HinTak:
You don't need to use dependency walker if you are cross-compiling.

```
winedump -j import <freetype.dll> | grep offset
```

does the job on the typical Linux/Mac machines too, if you have wine installed.

madig:
Okay, it builds and the smoke test works :) Now to make it nice.

madig:
@anthrotype The Appveyor log has the following:
```
C:\Python27\lib\site-packages\wheel\pep425tags.py:75: RuntimeWarning: Config variable 'Py_DEBUG' is unset, Python ABI tag may be incorrect
C:\Python27\lib\site-packages\wheel\pep425tags.py:79: RuntimeWarning: Config variable 'WITH_PYMALLOC' is unset, Python ABI tag may be incorrect
C:\Python27\lib\site-packages\wheel\pep425tags.py:85: RuntimeWarning: Config variable 'Py_UNICODE_SIZE' is unset, Python ABI tag may be incorrect
C:\Python36-x64\lib\site-packages\wheel\pep425tags.py:75: RuntimeWarning: Config variable 'Py_DEBUG' is unset, Python ABI tag may be incorrect
C:\Python36-x64\lib\site-packages\wheel\pep425tags.py:79: RuntimeWarning: Config variable 'WITH_PYMALLOC' is unset, Python ABI tag may be incorrect
```
Has this been an issue in your experience?

anthrotype:
you can ignore those warnings

madig:
Ok :)

Hm, from sifting through the vcpkg files for FreeType (https://github.com/Microsoft/vcpkg/tree/master/ports/freetype), they build FT without Harfbuzz, too. You'd also need to maintain custom port files I guess to keep the version number under control and not install zlib, etc. by default. From a cursory glance, this would currently amount to a vcpkg fork or at least clone of the repo plus a different --vcpkg-root.

ras0219-msft:
We'd be happy to accept PRs to make these optional (zlib, bzip2, png) so you can use the upstream copies if you'd like -- we've recently done something similar with OpenCV.

That said, a fork is 100% the way for you to have maximum control over your dependencies including build flags/versions/patches etc.

madig:
Hi! Forking is precisely what I'd like to avoid, as it would be an awful
lot of effort over just downloading a tarball and relying on MingW (which
is preinstalled on Appveyor anyway) plus hacks. I like the idea of vcpkg
however, is supporting set versions on the to-do list? Or just pointing
vcpkg at a dir with the build files so one does not have to fork the entire
repo to get custom stuff built? What about recursive build requirements
(build freetype without harfbuzz first, build harfbuzz, rebuild freetype
with harfbuzz support)?

anthrotype:
vcpkg is good, but here we don't need to use MSVC to build our embedded DLL, as this is not a normal CPython extension module that needs to be linked with the python DLL (and hence must be compiled with the same MSVC compiler version that python itself was built with). This is standalone Windows DLL that is loaded dynamically via ctypes. Building the DLL with mingw-w64 gcc under msys2 environment and bundling all the dependencies statically inside it already works fine for our use case.

ras0219-msft:
@madig If all you need is stability (i.e. once you have something working, you don't want us to automatically upgrade underneath you and break everything) then it is as simple as checking out the same commit from our upstream vcpkg repository. Each commit has hashes for all downstream libraries, so you're guaranteed to get exactly the same results every time, as long as you keep pulling the same tarball (for example: https://github.com/Microsoft/vcpkg/archive/a477fd6b6eb5a4053770e50423c6f1144c338c4d.zip).

If forking sounds harder than having a few files you replace after extracting, that's fine too! We build whatever is in the `ports\` directory when you run `vcpkg install`, so you can alter those freely.

The harfbuzz-freetype situation is quite challenging for every package manager it seems (which is why we haven't tackled it). My gut says that the practical way to solve it is to merge the two into a single "harfbuzz-freetype" build which internally does the loop. You could replace our existing `harfbuzz` and `freetype` ports with dummy forwarders to the combo-build, and all the other scripts should just work :).

madig:
Thanks for the pointers :) What I'd like to see is to make vcpkg take a file:// URL, so that I could in a CI script say `vcpkg install file://./ci/freetype` or something (analoguous to how Python's pip can look up PyPI but also takes a local directory), without messing with the existing ports or downloading the entire repo in addition to everything else for each commit. That would really simplify matters :) And I guess it would make it easier for other developers who need to pin versions.

I'll see what I can contribute to vcpkg once I have this figured out for this project.

madig:
Currently, `pip install <freetype-py directory>` installs without bundling even when the env var is set. I guess that's okay?

anthrotype:
Why?

anthrotype:
pip install <local folder> should be the same as cd-ing to that folder and running setup.py install. Pip actually copies the whole directory to a temporary location and runs setup.py from there.

madig:
No idea. Will investigate after I get this stuff to build again 😕

madig:
Hrm, https://docs.travis-ci.com/user/caching/#Things-not-to-cache says I shouldn't cache things that are slow to download but quick to install because no speed-gains are to be had, so, tarballs I guess? :(

And what the hell is wrong with the Windows build...

madig:
Apparently to generate the def file for the shared object or something...

madig:
I can reproduce the failure locally when I have `mingw-w64-$MSYS_ARCH-python(2|3)` installed in my MSYS2 environment. Maybe it's a PATH problem. Which reminds me of the PATH-manipulation of setup.py... Let's see.

madig:
This is... irritating. I suppose I'd be able to use native tooling when using CMake instead?

rougier:
I just discovered this long discussion and I'm several PR back. @madig @anthrotype can you tell me what's the status of this PR ? Maybe at this point it would be safer to create an experimental branch such that you can push your changes and set travis properly, what do you think?

madig:
Hi! This is my experimental branch, don't merge it ;)

I'm basically fighting compilation on Windows, Travis with Linux and macOS
I know how to get working. Once I'm done, the deployment step has to be
filled in by you and then we're done :)

Appveyor has been costing me quite some hair this week though and I'm out
of ideas currently.. except redoing everything with Python plus cmake
instead of msys etc...

anthrotype:
> Maybe it's a PATH problem. Which reminds me of the PATH-manipulation of setup.py.

yes. in which folder is that msys2 python installed? Is that present in the $PATH of the `bash` shell that is spawned from the setup.py?

madig:
IT WORKS 🍾 🎆 

MSYS2 comes with a MingW-Python2 installed, which sits in `/mingw(32|64)/bin` (top of the PATH var) and therefore is used by Harfbuzz' gen-def.py file through the shebang line (`#!/usr/bin/env python`). Here, a path mixup happens: PATH is in Windows format, but the headers get passed in in Unix format (MSYS2's path heuristics fail?), so the Python interpreter tries to open Unix paths while thinking of Windows and gets confused. This is solved by symlinking the MSYS2 Python2 into `/mingw(32|64)/bin`.

madig:
So... about shared libraries. I want to build Harfbuzz alongside FreeType -- how should they be distributed? Two separate libraries that both get included in the wheel? Or one big shared library made from duct-taping two static libraries together?

HinTak:
FWIW, that's exactly how the windows and mac os x FreeType backends of FontVal are built - they are built as shared/dynamics libraries against static library dependencies. I don't think they include harfbuzz (since FontVal does not need harfbuzz specific functionalities), but static versions of libpng and libbz2 are included, in exactly the way I described.

madig:
Ah, okay. I was confused by the duct-taping I had to do for Windows. Building a shared FreeType library on Windows still needs MingW but oh well.

HinTak:
Argh. I choose not to include harfbuzz because harfbuzz itself in turn requires gnulib and some other stuff. But in principle, you should be able to make shared libraries against static ones, if you have further dependencies all in static versions also.

gcc (including mingw) prefers dynamic libraries over static libraries if both are available. So the trick is moved the dynamic libraries aside (or copy/link the static libraries elsewhere). e.g. something like this:

i686-w64-mingn32-gcc -shared -o freetype.dll *.o -L<some_place_else> -lharfbuzz -lotherstuff1 -lotherstuff2 ...

where <some_place_else> is populated with static versions only of the dependent libraries, or symlinks of them.

madig:
I just found https://github.com/Microsoft/vcpkg/tree/master/ports/freetype (see my freetype-devel post). I imported the 0001 patch.

The build worked without MingW, but thanks for the info Hin-Tak, you never know when you're going to need it again :)

So, the sequence of commands for using VS and CMake on Windows is:

1. Download and unpack FreeType and Harfbuzz tarballs
2. cd FreeType-dir
    1. `mkdir build; cd build`
    2. Import 0001 patch
    3. `cmake -DWITH_HarfBuzz=OFF -DWITH_PNG=OFF -DWITH_BZip2=OFF -DWITH_ZLIB=OFF -DCMAKE_INSTALL_PREFIX="C:\FullPath\tmp\prefix\" ..`
    4. `cmake --build . --config Release --target install`
3. cd Harfbuzz-dir
    1. `mkdir build; cd build`
    2. `cmake -DHB_HAVE_FREETYPE=ON -DHB_HAVE_GLIB=OFF -DCMAKE_PREFIX_PATH="C:\FullPath\tmp\prefix\" -DCMAKE_INSTALL_PREFIX="C:\FullPath\tmp\prefix\" ..`
    3. `cmake --build . --config Release --target install`
4. cd FreeType-dir/build
    1. `cmake -DBUILD_SHARED_LIBS:BOOL=true -DWITH_HarfBuzz=ON -DWITH_PNG=OFF -DWITH_BZip2=OFF -DWITH_ZLIB=OFF -DHARFBUZZ_INCLUDE_DIRS="C:\FullPath\tmp\prefix\include\harfbuzz" -DCMAKE_PREFIX_PATH="C:\FullPath\tmp\prefix\" -DCMAKE_INSTALL_PREFIX="C:\FullPath\tmp\prefix\" ..`
    2. `cmake --build . --config Release --target install`

This will create a `freetype.dll` in `C:\FullPath\tmp\prefix\bin`. :)

madig:
Okay, I'm preparing a Python build script thing so we don't need a POSIX environment at all for building. What I currently have:
```Python
import os
import subprocess
import tarfile

try:
    import urllib.request
except ImportError:
    import urllib

FREETYPE_HOST = "https://download.savannah.gnu.org/releases/freetype/"
FREETYPE_TARBALL = "freetype-2.9.tar.bz2"
FREETYPE_URL = FREETYPE_HOST + FREETYPE_TARBALL
HARFBUZZ_HOST = "https://www.freedesktop.org/software/harfbuzz/release/"
HARFBUZZ_TARBALL = "harfbuzz-1.7.5.tar.bz2"
HARFBUZZ_URL = HARFBUZZ_HOST + HARFBUZZ_TARBALL


def cmd(string, cwd=None):
    subprocess.run(string, shell=True, check=True, cwd=cwd)


def download(url, target_path):
    try:
        from urllib.request import urlretrieve
    except ImportError:
        from urllib import urlretrieve

    print("Downloading {}...".format(url))
    urlretrieve(url, target_path)


root_dir = "."
prefix_dir = os.path.abspath(os.path.join(root_dir, "local"))
os.makedirs(prefix_dir, exist_ok=True)
build_dir = os.path.join(root_dir, "build")
build_dir_ft = os.path.join(build_dir,
                            FREETYPE_TARBALL.split(".tar")[0], "build")
build_dir_hb = os.path.join(build_dir,
                            HARFBUZZ_TARBALL.split(".tar")[0], "build")
os.makedirs(build_dir, exist_ok=True)
os.makedirs(build_dir_ft, exist_ok=True)
os.makedirs(build_dir_hb, exist_ok=True)

ft_tarball = os.path.join(build_dir, FREETYPE_TARBALL)
if not os.path.exists(ft_tarball):
    download(FREETYPE_URL, ft_tarball)

    with tarfile.open(ft_tarball, 'r:bz2') as tb:
        tb.extractall(build_dir)

hb_tarball = os.path.join(build_dir, HARFBUZZ_TARBALL)
if not os.path.exists(hb_tarball):
    download(HARFBUZZ_URL, hb_tarball)

    with tarfile.open(hb_tarball, 'r:bz2') as tb:
        tb.extractall(build_dir)

os.putenv("CMAKE_PREFIX_PATH", prefix_dir)

print("\n# First, build FreeType without Harfbuzz support")
cmd("cmake -DBUILD_SHARED_LIBS:BOOL=false "
    "-DWITH_HarfBuzz=OFF -DWITH_PNG=OFF -DWITH_BZip2=OFF -DWITH_ZLIB=OFF "
    "-DCMAKE_INSTALL_PREFIX=\"{}\" ..".format(prefix_dir),
    cwd=build_dir_ft)
cmd("cmake --build . --config Release --target install", cwd=build_dir_ft)

print("\n# Next, build Harfbuzz and point it to the FreeType we just build.")
cmd(
    "cmake -DBUILD_SHARED_LIBS:BOOL=false "
    "-DCMAKE_POSITION_INDEPENDENT_CODE=ON "  # Needed on Linux for next step.
    "-DHB_HAVE_FREETYPE=ON -DHB_HAVE_GLIB=OFF "
    "-DCMAKE_INSTALL_PREFIX=\"{}\" ..".format(prefix_dir),
    cwd=build_dir_hb)
cmd("cmake --build . --config Release --target install", cwd=build_dir_hb)

print("\n# Lastly, rebuild FreeType, this time with Harfbuzz support.")
cmd("cmake -DBUILD_SHARED_LIBS:BOOL=true -DMINGW=ON "
    "-DWITH_HarfBuzz=ON -DWITH_PNG=OFF -DWITH_BZip2=OFF -DWITH_ZLIB=OFF "
    "-DSKIP_INSTALL_HEADERS=ON "
    "-DHARFBUZZ_INCLUDE_DIRS=\"{}\" "
    "-DCMAKE_INSTALL_PREFIX=\"{}\" ..".format(
        os.path.join(prefix_dir, "include", "harfbuzz"), prefix_dir),
    cwd=build_dir_ft)
cmd("cmake --build . --config Release --target install", cwd=build_dir_ft)
```
Works on Windows (32bit build only so far) and Ubuntu 16.04. Still need to test on macOS (and instruct it make fat binaries).

I think I can more or less drop this into setup.py and have something nice and comprehensive :)

anthrotype:
Maybe you may find this useful
https://github.com/wingtk/gvsbuild

madig:
Hm. Not too bad for a first shot into the blue :)

madig:
Yay :)

Now it's just the Windows build left.

madig:
Now it's just the 32bit build *head scratching*

madig:
Yes! Travis and Appveyor succeeded :) 

https://travis-ci.org/rougier/freetype-py/builds/351559575
https://ci.appveyor.com/project/madig/freetype-py/build/1.0.66

Interesting how the Windows DLLs are about 500 KB in size while the Linux builds are around 2 MB (and the fat macOS binaries about 4 MB). Using `nm` on the Linux SOs prints public _and_ private functions... Gotta investigate what exactly gets exported/linked.

@rougier Get ready to fill in the deployment step :)

madig:
Note to self: find out if setup.py can be modified such that I don't need an env var for deciding whether to bundle or not.

madig:
Meh. The CMake-generated FT library with the static Harfbuzz lib is huge, like 2MB stripped of debug symbols, compared to producing the same binary with Autotools (~ 770 KB). On top of that, the external symbol table is filled with seemingly all symbols of Harfbuzz, while the Autotools includes just a few as non-external symbols. Autotools furthermore passes the linker an explicit symbol export file, which I can't seem to replicate in a platform-agnostic way on CMake...

madig:
@rougier I'm mostly done with this PR I think. Have a look :) To pull it in officially you need an Appveyor account and you need to set up the deployment details.

I'll try to resolve the symbol and size difference stuff in upstream FT and HB.

setup.py
```diff
@@ -4,34 +4,148 @@
...
+            ) + bdist_wheel.get_tag(self)[2:]
+
+    cmdclass['bdist_wheel'] = UniversalBdistWheel
+
+
+class SharedLibrary(Extension):
+    """Object that describes the library (filename) and how to make it.
+
+    Note: The `output_dir` parameter shouldn't include the final lib/ or bin/
+    component, we figure that out on the fly.
```
anthrotype: I don't think this class should also determine whether the generated output goes to "lib" or "bin", or the prefix of the library. It's better to leave it agnostic as to the build system used. The library file extension is a different thing, as that is always the same (depending on platforms), and is used in get_ext_filename method to override the default extension module filename.

anthrotype: (i like to make this into a reusable setuptools extension which one could use as a build requirement at some point; better not to add too much freetype-py specific logic to it and keep it general)

setup.py
```diff
@@ -4,34 +4,148 @@
...
+        self.cwd = path.normpath(cwd)
+        self.output_dir = path.join(path.normpath(output_dir), self.lib_dir)
+        self.env = env or dict(os.environ)
+
+
+class SharedLibBuildExt(build_ext):
+    """Object representing command to produce and install a shared library."""
+
+    # Apparently needed for determining the correct library filename on
+    # Windows.
```
anthrotype: remove this comment. that's not only needed for windows; it's the thing that makes it possible to make setuptools and wheel to believe that this module contains a native extension module (whereas in reality it contains a shared library).

setup.py
```diff
@@ -4,34 +4,148 @@
...
+
+
+cmdclass['build_ext'] = SharedLibBuildExt
+
+if os.environ.get("FREETYPEPY_BUNDLE_FT"):
+    print("# Will build and bundle FreeType.")
+    ext_modules = [
+        SharedLibrary(
+            "freetype.freetype",  # double name and dot are needed for macOS.
+            cmd="python ./setup-build-freetype.py",
```
anthrotype: why not call the script just "build_freetype.py"?

anthrotype: also, you should use sys.executable, not "python", so it will execute the script with the same python interpreter used to run the setup.py

setup.py
```diff
@@ -4,34 +4,148 @@
...
+            dry_run=self.dry_run)
+
+
+cmdclass['build_ext'] = SharedLibBuildExt
+
+if os.environ.get("FREETYPEPY_BUNDLE_FT"):
+    print("# Will build and bundle FreeType.")
+    ext_modules = [
+        SharedLibrary(
+            "freetype.freetype",  # double name and dot are needed for macOS.
```
anthrotype: remove the comment. this is not just needed for macOS.
This SharedLibrary class should look like the Extension class, so the first argument is the name of the module, including the package where the module is located.
It should be "freetype.libfreetype".

Note, it doesn't matter the actual file name of the shared library, because this one is embedded and we load it with its full name via ctypes. It's ok to have "lib" as prefix in the filename for all three platforms for simplicity (including windows where it would not commonly be there when built with msvc).

setup.py
madig: I did that at some point and tox tripped over somehow.. but gonna try again.

I suppose I can shorten the name. Just looks nicer next to setup.py ;)

anthrotype: > tox tripped over somehow

try passing to subprocess a list of strings, instead of a single string

rougier:
@madig Thanks. I just activated the AppVeyor but I'm not sure about deployment details.

madig:
I suppose you fill in login-information for PyPI in .travis.yml and appveyor.yml in deploy/on_success. Have a look at both files in https://github.com/fonttools/ttfautohint-py.

anthrotype:
You’ll have to encrypt them first! You don’t want to share them as plain text. I’m away now (typing from my phone).

anthrotype:
Relevant docs here:

https://docs.travis-ci.com/user/deployment/pypi/

https://www.appveyor.com/docs/build-configuration/#secure-variables

rougier:
He he, not at all 😄. I first need to test the PR locally and then we'll see for Travis/AppVeyor (I'm not so familiar with them as you may have guessed)

rougier:
Just tested it, works like a charm on OSX.

The README needs some formatting and the first described option is not that clear im my opinion. I first thought you were implying the default install would come with pre-compiled binairies.

Is there a way to use Travis and/or AppVeyor on a PR ?

madig:
There is, and you're looking at it :) When you set Travis and Appveyor on a repo, both services will also look at PRs -- if the branch the PR points at has a `.travis.yml` and `appveyor.yml` file, both servies will automatically run them on the PR, continuously :)

Alright, will reword the Readme some. Do you use a rst-formatter?

rougier:
I see the travis check but not the app veyor. Does it need the manual webhook installation ?

For the README, it not really urgent, we can do that later. And rst is fine.

madig:
No idea. Look if you can find Appveyor in https://github.com/rougier/freetype-py/settings/installations -- Travis shows up as an app, but I can't find Appveyor in my repo 😕

rougier:
Not clear. I configures AppVeyour to listen to PR but yours does not appear on AppVeyor board. I'm investigating.

madig:
Squashed all commits into one. This includes reverting src/freetype to freetype/.

README.rst
```diff
@@ -1,54 +1,67 @@
...
+------------------------------------------------
 
-Mac users (third way)
----------------------
+If you don't want to or can't use the pre-built binaries, build FreeType
+yourself: `export FREETYPEPY_BUNDLE_FT=yesplease && python setup.py install`.
+This will download and compile FreeType with Harfbuzz support as specified in
+`setup-build-freetype.py`. Set the environment variable `PYTHON_ARCH` to 32 or
+64 to explicitly set an architecture, default is whatever your host machine
+uses. On macOS, we will always build a 96 bit binary.
```
anthrotype: Never heard “96 bit”. You mean the “fat” 32/64 bit binary?

setup-build-freetype.py
```diff
@@ -0,0 +1,166 @@
...
+prefix_dir = path.abspath(path.join(build_dir, "local"))
+build_dir_ft = path.join(build_dir, FREETYPE_TARBALL.split(".tar")[0], "build")
+build_dir_hb = path.join(build_dir, HARFBUZZ_TARBALL.split(".tar")[0], "build")
+
+CMAKE_GLOBAL_SWITCHES = ("-DCMAKE_COLOR_MAKEFILE:BOOL=false "
+                         "-DCMAKE_PREFIX_PATH:PATH=\"{}\" "
+                         "-DCMAKE_INSTALL_PREFIX:PATH=\"{}\" ").format(
+                             prefix_dir, prefix_dir)
+
+bitness = None
```
anthrotype: Where is this “bitness” variable used?

setup.py
```diff
@@ -4,34 +4,136 @@
...
+
+
+cmdclass['build_ext'] = SharedLibBuildExt
+
+if os.environ.get("FREETYPEPY_BUNDLE_FT"):
+    print("# Will build and bundle FreeType.")
+    ext_modules = [
+        SharedLibrary(
+            "freetype.libfreetype",  # package.shared_lib_name
+            cmd="\"{}\" ./setup-build-freetype.py".format(sys.executable),
```
anthrotype: You know you could use single quotes so that you don’t need to escape inner double quotes.

setup-build-freetype.py
madig: Further down in step 3 when deciding whether to turn on `-fPIC` or not.

setup.py
madig: `cmd.exe` doesn't like single quotes from what I gathered.

README.rst
madig: Yes 😁

setup-build-freetype.py
anthrotype: Oh I see, thanks

README.rst
anthrotype: I think it’s better to avoid that in the Readme.

setup.py
anthrotype: I meant to wrap the python string literal

HinTak:
Ditto... my home machine still have a tri-arch compiler to do 32-bit ppc + 32-bit intel + 64-bit intel fat binaries. They are just called tri-arch fat binaries...
(or bi-arch...).

README.rst
```diff
@@ -1,47 +1,66 @@
...
+high-level API is bound.
 
 Documentation available at: http://freetype-py.readthedocs.org/en/latest/
 
 Installation
 ============
 
-To be able to use freetype python, you need the freetype library version 2
-installed on your system.
+**From PyPI, recommended**: `pip install freetype-py`. This will install the
```
anthrotype: Note that on PyPI, you will also upload a sdist (.tar.gz or .zip) in addition to the precompiled wheel packages.
That will be used for architectures or python versions for which pip cannot find a suitable wheel, or if the user passes the --no-binary option. When installing from sdist, the value of the FREETYPEPY_BUNDLE_FT environment variable will determine whether the embedded shared lib will be built or not.

README.rst
```diff
@@ -1,47 +1,66 @@
...
+library with a bundled FreeType binary, so you're ready to go on Windows,
+macOS and Linux (all with 32 and 64 bit support).
+
+Installation with compiling FreeType from source
+------------------------------------------------
+
+If you don't want to or can't use the pre-built binaries, build FreeType
+yourself: `export FREETYPEPY_BUNDLE_FT=yesplease && python setup.py install`.
```
anthrotype: `pip install .` is the currently recommended method for installing from a source folder. The `setup.py install` method uses setuptools to create and install a (now deprecated) `egg` package, whereas `pip install .` will make a temporary `wheel` and install that with pip.
`pip install .` is also more future proof because build-system agnostic. In the future, one may use a different build tool than setuptools, and the next pip version will allow to specify custom build requirements.

setup.py
```diff
@@ -4,34 +4,136 @@
...
+            dry_run=self.dry_run)
+
+        distutils.file_util.copy_file(
+            lib_fullpath,
+            dest_path,
+            verbose=self.verbose,
+            dry_run=self.dry_run)
+
+
+cmdclass['build_ext'] = SharedLibBuildExt
```
anthrotype: You should pass an empty cmdclass dict if FREETYPEPY_BUNDLE_FT is not set. You don’t want to override the default bdist_wheel or build_ext commands in the case you are not compiling the embedded shared library.

madig: Should cmdclass include just `cmdclass['bdist_wheel'] = UniversalBdistWheel` even if not bundling FreeType?

anthrotype: Correct, leave it empty.

anthrotype:
this appveyor bot is noisy, isn't it? what info does it add besides the built-in github status checks?
Maybe it'd be good to disable it somehow.

madig:
@rougier you might want to tell AppVeyorBot to not post on PRs because it's just noise ;)

madig:
This testing for what works where is tiring...

madig:
No idea if this actually does what I want.

anthrotype:
nice! 👍 
do you know know it's possible to use pip to install both cmake and ninja?
somebody at Kitware packaged them up as python wheels and uploaded to PyPI :)

madig:
```
$ python3 -m venv venv
$ . venv/bin/activate
$ pip install cmake ninja
Collecting cmake
  Downloading https://files.pythonhosted.org/packages/19/aa/ab833932dbf8e0c56eae2bf5896c3a18afc600c77286f0bd8ff809a35dc9/cmake-0.9.0-cp36-cp36m-manylinux1_x86_64.whl (20.2MB)
    100% |████████████████████████████████| 20.2MB 76kB/s 
Collecting ninja
  Downloading https://files.pythonhosted.org/packages/fe/15/c6c9689f00b69ec8ebf8fa7b357be7a84a65de8cd9995f5b887ede1de0e6/ninja-1.7.2.post2-cp36-cp36m-manylinux1_x86_64.whl (89kB)
    100% |████████████████████████████████| 92kB 3.5MB/s 
Installing collected packages: cmake, ninja
Successfully installed cmake-0.9.0 ninja-1.7.2.post2
$ which cmake
/tmp/venv/bin/cmake
$ which ninja
/tmp/venv/bin/ninja
```
Looks like it. Although you have to download a 20 MB cmake. Not sure if that's faster than just using the default...

I'm currently thinking of updating the minimum cmake vrsion FreeType's CMakeLists.txt requires, but the Manylinux1 container used here is a CentOS 5.11 image, which only has 2.6 in the repos. Harfbuzz requires 2.8 and it obviously works fine though, so... mh.

anthrotype:
one could add `$HOME/.cache/pip` to the `cache` directories in .travis.yml, to avoid redownloading the same wheels
https://docs.travis-ci.com/user/caching#pip-cache

madig:
Hm yeah. Gotta look up Appveyor caching again...

So I installed Docker and launched the manylinux1 image. Turns out that CMake 2.8.11.2 is installed by default, hooray! Ninja doesn't exist in the repos, though.

HinTak:
"freetype library version 2" sounds a bit odd - nobody had used freetype version 1 for about 20 years...

madig:
I just stumbled over https://pypi.python.org/pypi/auditwheel, which can graft shared libs onto a wheel so you don't have to build things statically. Something to investigate some other time.

anthrotype:
Yeah, I think multibuild should run that automatically for you already. But we don’t need it here, as we managed to bundle everything into one shared library.

HinTak:
I thought it would be useful to suggest logging the actual version of freetype used in any kind of testing (especially if you are using the system one). You can do it with:

```
import freetype
print freetype.version()
```

(python 2 syntax for print).

I had to do that in the last few days hunting down a bug, because of difference in the two machines I use - I have the latest on my main machine, but WSL (i.e. Ubuntu ?) on windows 10 have freetype 2.6.1 which is rather antique. It also rendered rather differently, because of the changes in the byte-code interpreter during late 2.6.x . I have to explicitly set the interpreter version to get them to agree.

madig:
Do we need the duplication of installation instructions in the README and Introduction?

tests/tests/test_ft_face.py
```diff
@@ -1,6 +1,11 @@
...
+
 import freetype
 
 
 def test_ft_face():
     """A smoke test."""
     assert freetype.Face('../examples/Vera.ttf')
+
+    if os.environ.get("FREETYPEPY_BUNDLE_FT"):
+        assert freetype.version() == (2, 9, 0)
```
anthrotype: then you'll need to remember to update this test every time your bump the bundled freetype shared library

madig: Hm. You mean I should scrape this from setup-build-freetype.py? :grin:

anthrotype: yeah, you could do that to DRY

hey I just noticed, why do you have two nested `tests/tests/` subfolders?

madig: Because @rougier was against placing freetype inside a Lib/ folder ;)

anthrotype: that's horrible. you don't need that. pytest can just find the tests wherever they are in current folder and all its subfolders...

madig: Yes, but I want tox-safeness :syringe:

anthrotype: I still don't get it. what's the problem with cd'ing into the tests directory (where .py test files are) and run pytest from there?

madig: Dunno, but if that ensures that the installed package is toxed, okay :grin:

anthrotype: but you've already set `changedir=tests` here in tox.ini, no?
https://github.com/madig/freetype-py/blob/d46e6347a9b9c782de8631d5d6d936bfbb9d7ab8/tox.ini#L7

that should already ensure that the pytest command is executed with a current working directory set to "tests", and thus free from the stuff which is in the root of the repo (setup.py, the freetype package source itself, etc.)

anthrotype: I just tried locally to `git mv tests/tests/test_ft_face.py tests/test_ft_face.py` and run `FREETYPEPY_BUNDLE_FT=1 tox` and it just worked.

By the way, I use [`tox-venv`](https://github.com/tox-dev/tox-venv) instead of plain `tox` on my machine. Tox-venv uses the built-in `venv` module (python3 only) whereas `tox` still uses the `virtualenv` module (with all its hacks to make it work for both 2 and 3), though they are discussing about making tox-venv the default for python3.

Anyway, when you `python3 -m venv somenthing`, it does not automatically install the `wheel` package for your, unlike the old virtualenv. Your setup.py raise an import error when attempting to build the embedded shared library if wheel is not installed in the current environment.
You should probably add `wheel` in the `deps=` list inside tox.ini, or make it optional in setup.py.

madig: Brain lapse, will implement :)

Tox runs both py27 and py36 tests fine, but will add wheel.

Now for how to scrape the version information...
```
    def test_bundled_version():
        if os.environ.get("FREETYPEPY_BUNDLE_FT"):
>           from .. import setup_build_freetype
E           ValueError: Attempted relative import in non-package
```

anthrotype: Don’t import it, open the file and regex it

madig:
@rougier What's keeping you? :)

tests/smoke_test.py
```diff
@@ -0,0 +1,24 @@
...
+import pytest
+
+
+def test_load_ft_face():
+    """A smoke test."""
+    assert freetype.Face('../examples/Vera.ttf')
+
+
+def test_bundle_version():
+    if os.environ.get("FREETYPEPY_BUNDLE_FT"):
```
anthrotype: hm, that environment variable is only needed at setup time, should not change anything when running tests. 
to detect if running freetype-py with a bundled library you should rather check that there is one at the expected path.

madig: What if for whatever reason a `.so` isn't present when it should? Testing for the environment variable makes it clear that there must be something.

anthrotype: > What if for whatever reason a .so isn't present when it should?

then something went wrong in the installation and should fail!

> Testing for the environment variable makes it clear that there must be something.

the env variable isn't required at runtime (and hence, at test time), only when you run setup.py.

anthrotype: just skip the test if it's not present

madig: I mean, is it enough to os.path.exists(glob.glob(packagedir / libfreetype*)) in the test? I suppose setup will fail if the env var is set but no .so is put anywhere, but what if it doesn't under some circumstance?

setup-build-freetype.py
```diff
@@ -42,7 +42,9 @@
                              prefix_dir, prefix_dir)
 
 # Try to use Ninja to build things if it's available. Much faster.
-if distutils.spawn.find_executable("ninja"):
+# On Windows, I first need to figure out how to make it aware of VC, bitness,
+# etc.
```
anthrotype: what was the issue with ninja on windows? cmake should tell ninja everything it needs to know.
what versions of cmake/ninja are you using?

madig: cmake 3.10.x and the most up-to-date Ninja I could download. Last time I tried, it failed to find a compiler... maybe just a path problem, but I wanted to build a bwheel and worry about Ninja later. Does it work for you?

anthrotype: If I run setup.py from within the "Visual C++ 2015 x64 Native Build Tools Command Prompt", it works

anthrotype: well, actually if fails later when linking with unresolved symbols.. argh. Sorry can't debug this further.

madig: I want it to run from a normal Powershell without "* Command Prompt" frickery... some other time. msbuild gets the job done somehow.

anthrotype:
@madig for some reason, on my Windows 10 VM running Python 3.6.4 64-bit, I was getting `SSL: CERTIFICATE_VERIFY_FAILED` errors when the setup-build-freetype.py script attempts to download the freetype and harfbuzz tarballs. I have no idea why, but I was able to work around it by pip installing `certifi` and passing it to `urlopen`.
Is it downloading fine for you on Windows?

```diff
diff --git a/setup-build-freetype.py b/setup-build-freetype.py
index d9a879d..cd87acb 100644
--- a/setup-build-freetype.py
+++ b/setup-build-freetype.py
@@ -95,12 +95,18 @@ def shell(cmd, cwd=None):
 def download(url, target_path):
     """Download url to target_path."""
     try:  # Python2 compatibility wrapper.
-        from urllib.request import urlretrieve
+        from urllib.request import urlopen
     except ImportError:
-        from urllib import urlretrieve
+        from urllib2 import urlopen
+    try:
+        import certifi
+    except ImportError:
+        certifi = None

     print("Downloading {}...".format(url))
-    urlretrieve(url, target_path)
+    with urlopen(url, cafile=certifi.where() if certifi else None) as resp:
+        with open(target_path, "wb") as f:
+            f.write(resp.read())


 def ensure_downloaded(url, sha256_sum):

```

anthrotype:
just realised that `with urlopen(...) as response` won't work on python2, so you can either call `.close()` manually or use the `contextlib.closing` context manager

madig:
It's downloading fine for me... but we do happen to have certifi installed. Huh.

madig:
Note to self: look at https://gist.github.com/roxlu/0108d45308a0434e27d4320396399153, salvage useful things

rougier:
Sorry for the long silence, I was quite busy and I lost track of this PR. Is it ready to be merged ?

madig:
I think so.

Maybe the text in introduction.rst should refer to the readme?

You need to fill in the deployment stuff afterwards, though :)

anthrotype:
> You need to fill in the deployment stuff 

or you could add Nikolaus as collaborator on the PyPI admin page for the freetype-py project and he could set it up for you.
https://pypi.org/manage/project/freetype-py/collaboration/
@madig needs to tell @rougier what is his PyPI username (I presume `madig`?).

ci/multibuild_config.sh
```diff
@@ -0,0 +1,22 @@
+# Define custom utilities
+# Test for OSX with [ -n "$IS_OSX" ]
+
+function pre_build {
+    # Any stuff that you need to do before you start building the wheels
+    # Runs in the root directory of this repository.
+    if [ "$TRAVIS_OS_NAME" == "linux" ] && [ "$PLAT" == "i686" ]; then
+        apt install gcc-multilib g++-multilib
```
anthrotype: @madig why do you need to do `apt install gcc-multilib g++-multilib` inside the manylinux1 docker images? Firstly, these are CentOS based and thus they use `yum`, not `apt`. But also, multibuild already should pull the i686 version of the manylinux1 dockerfile automatically, which should have all the 32 bit toolchain already set up...

madig: Hm. The test below also doesn't make much sense as it's also run in manylinux1. The load test still passes so I guess it's using the installed FT.

madig:
I don't have a PyPI account. @rougier do you want to fill in the deployment stuff yourself or hand that off to someone?

@anthrotype Why are you using "skip_cleanup: true", "distributions: check" and "on: repo: ..." in ttfautohint?

anthrotype:
> Why are you using "skip_cleanup: true", "distributions: check" and "on: repo: ..." in ttfautohint? Shouldn't 'distributions: "sdist bdist_wheel"' suffice?

anthrotype:
> Why are you using "skip_cleanup: true", "distributions: check" and "on: repo: ..." in ttfautohint? Shouldn't 'distributions: "sdist bdist_wheel"' suffice?

because we have already built the wheels inside the multibuild environment, we don't want Travis dpl tool to try rebuiding them again outside of it, which it will attempt to do.
The default `distributions` value is `sdist bdist_wheel`; these are simply forwarded by Travis to the setup.py script. I set it to `check` so that it doesn't actually do anything. The `check` setuptools command just checks the package metadata is correct. In other projects, I defined a `pass` custom command that does nothing, for the same reasons.
The skip_cleanup is to prevent Travis from deleting the wheels we just built.
I know, it sounds complicated, but it works. 
The alternative would be to `pip install twine` and call twine ourselves in a custom deploy script, avoiding the use of Travis built-in PyPI intergration.

anthrotype:
it's to prevent the deploy stage to trigger from forks.

madig:
@rougier  ping :)

rougier:
Yes, sorry. I'm not too reactive lately (lot of work).  Is this ready to be merged ?

madig:
Yes, the only thing left to do is filling in the PyPI upload credentials. You can do that yourself or follow Cosimo's suggestion (https://github.com/rougier/freetype-py/pull/79#issuecomment-379685275).

rougier:
Ok, I will add them later then. Found how to do it on with Travis but not with apppveyor.

anthrotype:
For Appveyor, you need to encrypt your PyPI password by going to Account → Encrypt data. Store the secure value as TWINE_PASSWORD in appveyor.yml

rougier:
Just dit it. Travis still failing. We need to add a travis/appveyor status in the README byt the way.

rougier:
```
  File "ci/multibuild/supported_wheels.py", line 9, in <module>
    from pip.pep425tags import get_supported
ModuleNotFoundError: No module named 'pip.pep425tags'
ERROR: You must give at least one requirement to install (see "pip help install")
```

anthrotype:
Try update the multibuild submodule ti latest master, or even try the dev branch if it still fails

rougier:
It seems to be up to date and the PR was ok when not merged.

madig:
https://github.com/matthew-brett/multibuild/issues/159 -- seems to be this?

rougier:
Yes. Note sure how to switch to the devel branch on the module though. Maybe it's not worth the trouble.

madig:
You can do follow https://stackoverflow.com/questions/5828324/update-git-submodule-to-latest-commit-on-origin and use the commit hash of the latest devel branch.

PR #80, opened 2018-03-02
Move freetype/ to src/freetype/

madig:
... to apply the recommendations in
https://blog.ionelmc.ro/2014/05/25/python-packaging/, to make tox
testing work better.


anthrotype:
it's not just about tox or any specific test runner. It's about being able to explicitly run tests against an installed package, not what happens to be importable from the current directory, thus catching issues in the setup.
Though it does require a developer to do an extra editable install.

madig:
:)

Oh and, since I want to move away from submodules, moving it to src/ instead (no python/) is a bit neater. Will amend PR.

madig:
You can merge this one, unless you're waiting for something?

rougier:
I'm not sure to get the point on this.

madig:
It is explained in https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure -- the main concern here is avoiding implicit imports in testing and when otherwise working on the library. You instead just run tox and it will install the package in a test directory. That way, you immediately catch install failures like an unspecified encoding of a Readme file before pushing a release to PyPI :)

rougier:
Yes, but I'm not really convinced by the argument.  I would prefer to stick to the current structure if you don't mind. We can change it later.

HinTak:
Actually for the sake of argument, to avoid implicit imports, you could just as well put your test scripts and the whole of test related infra-structures script into a sub-directory.

What you want, is that the directory in which you run your test, does not have a (mostly functional) "freetype" freetype-py module in it. You could also put your test harness within a sub-directory to achieve that purpose too.

The source code is going to need changing more often than the test scripts. My alternative proposal - i.e. putting the less-often changed test-scripts stuff in a sub-directory, is more attractive than what this change is proposing. This change requires developers to go into a directory every time before they start doing anything, and make it inconvenient for a more frequent operation.

My 2$...

madig:
And what about

> Without src you get messy editable installs ("setup.py develop" or "pip install -e"). Having no separation (no src dir) will force setuptools to put your project's root on sys.path - with all the junk in it (e.g.: setup.py and other test or configuration scripts will unwittingly become importable).

But okay, will keep to current structure...

PR #81, opened 2018-03-02
Add simple .travis.yml for running tox on everything

madig:


madig:
Ugh. Travis' macOS Python support seems to require workarounds to use. Let's drop that for now. The Travis stuff probably has to be rewritten for building wheels anyway.

anthrotype:
multbuild also takes care of downloading and installing the various pythons on macOS.

However running just tox won’t work with the proposed PR for building the embedded shared library from the git submodule, as I explained there (tox —develop would work though).

Also it’s more complicated than that, because for the Linux wheel we want to build it inside the manylinux1 docker (running some old CentOS) then testing it on a different Linux flavor (I think multibuild uses an Ubuntu Trusty image) to verify it works elsewhere.
So you’d have to do tox —installpkg $your_precompiled_wheel, e.g. see https://github.com/googlei18n/compreffor/blob/master/config.sh

madig:
Thanks for the info. This PR is more of a stop-gap.

PR #82, opened 2018-03-15
Expose FT_Outline_Decompose() as Outline.decompose(), with example

duelafn:
I wanted access to FT_Outline_Decompose(), so I added it in. See  examples/glyph-vector-decompose.py for example use.

I tried to follow existing conventions, I can fix things up if I made a blunder. Thanks for considering this PR!

### DIFF
--- a/examples/glyph-vector-decompose.py
+++ b/examples/glyph-vector-decompose.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# Copyright 2018 Dean Serenevy <dean@serenevy.net>
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+'''
+Show how to use outline decompose.
+'''
+import os
+import sys
+
+sys.path.append(os.path.dirname(__file__))
+sys.path.append(os.path.join(os.path.dirname(__file__), '..'))
+
+import freetype
+
+def move_to(a, ctx):
+    ctx.append("M {},{}".format(a.x, a.y))
+
+def line_to(a, ctx):
+    ctx.append("L {},{}".format(a.x, a.y))
+
+def conic_to(a, b, ctx):
+    ctx.append("Q {},{} {},{}".format(a.x, a.y, b.x, b.y))
+
+def cubic_to(a, b, c, ctx):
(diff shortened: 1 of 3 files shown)


rougier:
Hi @duelafn, sorry for the long delay in answering your PR. Many thanks, I think it'll be useful.
We're in the process of a big refactoring of the bindings (mostly to add the option of building the binary alongside the bindings). I would prefer to wait for the refactoring to occur before merging if it is fine for you (I think it won't change drastically you're PR but you might need to rebase at some point).

If you don't heard from me in the coming days, just ping me here (just in case I forgot)

duelafn:
Rebased, and thank you!

I've also coded up an alternate experiment where `decompose()` accepts an object instead of separate functions because. I wasn't trilled with passing/receiving a context in the original patch. However, I think that I like even less that the object approach requires methods named exactly right. I've published the object approach [in my "object-style" branch](https://github.com/duelafn/freetype-py/commit/226d2bb595fbf7abe27fa393d5ba89c852b79699) if you'd like to see it. I post it as a possible alternative to this patch in case you feel that it is a superior approach.

rastislav-chynoransky:
Nice feature :+1: Is it currently mergeable?

duelafn:
Merges cleanly for me and the example script still works in both python2 and python3.

rougier:
Thansk. Travsi is complaining but I'm not quite sure why.

anthrotype:
Try updating the multibuild git submodule

duelafn:
Ok, rebased again to get the latest multibuild. That fixed the tests.

rougier:
Cool. Thanks for your contribution.

loveencounterflow:
Notwithstanding any pending refactorings, is there any chance the general public (me) can gain access to a pip-installable intermediate release, as in v2.1post1 + `decompose()`?

madig:
You can install git master: https://stackoverflow.com/a/20101940

loveencounterflow:
Thanks a bunch, `python3.6 setup.py build` did work. 

I already managed to find the [first bug](https://github.com/rougier/freetype-py/issues/114) :-|

madig:
Well done 😁👍

PR #87, opened 2018-06-09
Update HarfBuzz in bundle to 1.8.0

madig:
Can you please make a new release soon so we get wheels on PyPI?


madig:
Where'd the Travis thing go? Hm. Maybe you need to re-register the project, there was a similar problem in fontbakery...

madig:
Ah, there it is :confused:

PR #89, opened 2018-06-29
solve exception when python2 with emoji(utf8) and ucs-2

Yangwoqizuo:
solve exception when python2 with emoji(utf8) and ucs-2


anthrotype:
Also check fonttools’ `byteord` function to solve the same problem for “narrow” python2 builds.
https://github.com/fonttools/fonttools/blob/ad597a49b4505233c2398642f1548d22c390a273/Lib/fontTools/misc/py23.py#L70

madig:
The library is now Python 3.6+ only.

PR #90, opened 2018-07-02
Unbreak CI wheel building on Linux

madig:
Edit: So, the 2.0 release was botched. The bundling didn't work on Linux because the FREETYPEPY_BUNDLE_FT variable didn't reach the Docker container. This PR makes it work. AppVeyor deployment may need a separate fix.

### DIFF
--- a/ci/custom_docker.sh
+++ b/ci/custom_docker.sh
@@ -1,0 +1,58 @@
+# Copy of upstream function with added FREETYPEPY_BUNDLE_FT environment var
+# passing.
+
+function build_multilinux {
+    # Runs passed build commands in manylinux container
+    #
+    # Depends on
+    #     MB_PYTHON_VERSION
+    #     UNICODE_WIDTH (optional)
+    #     BUILD_DEPENDS (optional)
+    #     DOCKER_IMAGE (optional)  
+    #     MANYLINUX_URL (optional)
+    #     WHEEL_SDIR (optional)
+    local plat=$1
+    [ -z "$plat" ] && echo "plat not defined" && exit 1
+    local build_cmds="$2"
+    local docker_image=${DOCKER_IMAGE:-quay.io/pypa/manylinux1_\$plat}
+    docker_image=$(eval echo "$docker_image")
+    retry docker pull $docker_image
+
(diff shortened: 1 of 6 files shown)


madig:
Something's wrong. The Linux bundling jobs do not bundle the library and the non-bundling job does not create a universal wheel. The macOS build seems to work as expected.

madig:
I forgot that the env var might not reach the test. Hm.

madig:
Slowly getting there... why did this not turn up before?!

madig:
Not really happy with creating a custom Docker wrapper for Linux, but creating an env_vars.sh file on the fly doesn't work... may change it again when I get feedback on https://github.com/matthew-brett/multibuild/issues/183.

ci/custom_docker.sh
```diff
@@ -0,0 +1,58 @@
...
+    #     UNICODE_WIDTH (optional)
+    #     BUILD_DEPENDS (optional)
+    #     DOCKER_IMAGE (optional)  
+    #     MANYLINUX_URL (optional)
+    #     WHEEL_SDIR (optional)
+    local plat=$1
+    [ -z "$plat" ] && echo "plat not defined" && exit 1
+    local build_cmds="$2"
+    local docker_image=${DOCKER_IMAGE:-quay.io/pypa/manylinux1_\$plat}
+    docker_image=$(eval echo "$docker_image")
```
belluzj: What does this line do?

madig: Dunno 😁

anthrotype:
I haven't deeply looked into this, but honestly, I don't like that you have to basically "fork" parts from multibuild to accomplish such a basic task of passing an environment variable to the docker container. We should ask multibuild maintainers how to best do this.

madig:
I did: https://github.com/matthew-brett/multibuild/issues/183

Will change once I hear from them.

anthrotype:
https://github.com/rougier/freetype-py/releases/tag/2.0.0.post1

for the title of the Github release notes, it think it's better to keep the same format as the previous ones: e.g. "Version 2.0.0.post1"

PR #93, opened 2018-07-03
Fix AppVeyor deployment script and add sdist

madig:
cmd.exe doesn't expand *.whl and newer twine versions work better with PowerShell, so get rid of the hack.

Also generate sdist on deploy.


anthrotype:
it's strange because the cmd hack with the glob pattern used to work, I remember using it for other projects. Anyway, don't have time to investigate, so this lgtm

madig:
Alright, here we go. post2!

PR #94, opened 2018-07-03
Fix AppVeyor deploy, take two

madig:


anthrotype:
it didn't work because the wheels are built in `dist` folder by default when you use the bdist_wheel comand, not in `wheelhouse`. Wrong copy-paste.

anthrotype:
keep the glob pattern and use `dist/*.whl` instead of `wheelhouse/*.whl`

madig:
Oh. Hm. How do I tell dpl to also upload the sdist then? In the deploy step (https://travis-ci.org/rougier/freetype-py/jobs/399574506) it only uploads the whl.

anthrotype:
no, it uploads anything inside the dist folder, trust me

.travis.yml
```diff
@@ -59,16 +59,16 @@ before_install:
...
 script:
     - install_run $PLAT
 
 after_success:
-  # copy compiled wheels to dist/ where Travis `dpl` tool can find them and 
+  # Copy compiled wheels to dist/ where Travis `dpl` tool can find them and 
   # upload to PyPI
   - if [ -n "$TRAVIS_TAG" ]; then mkdir -p dist; cp wheelhouse/*.whl dist; fi
+  # Additionally build sdist on tag and not bundling.
+  - if [ -z $"FREETYPEPY_BUNDLE_FT" && -n "$TRAVIS_TAG" ]; then python setup.py sdist; fi
```
anthrotype: again this is not valid. You need to move the `&&` outside the brackets, like so:

```sh
if [ -z "$FREETYPEPY_BUNDLE_FT" ] && [ -n "$TRAVIS_TAG" ]; then python setup.py sdist; fi
```

but you don't want to build the sdist for every build, only once. So export another variable and check that here

madig: Isn't that what's done here? It makes sure that 1. no bundling is going on 2. we are tagged?

anthrotype: because you're assuming that there is only one build that does not export FREETYPEPY_BUNDLE_FT variable; it may not always be like that, e.g. say we add another build to do linting, etc.

.travis.yml
```diff
@@ -67,8 +68,7 @@ after_success:
   # Copy compiled wheels to dist/ where Travis `dpl` tool can find them and 
   # upload to PyPI
   - if [ -n "$TRAVIS_TAG" ]; then mkdir -p dist; cp wheelhouse/*.whl dist; fi
-  # Additionally build sdist on tag and not bundling.
-  - if [ -z $"FREETYPEPY_BUNDLE_FT" ] && [ -n "$TRAVIS_TAG" ]; then python setup.py sdist; fi
+  - if [ -n "$BUILD_SDIST" ]; then python setup.py sdist; fi
```
anthrotype: you also need to check that it is a tagged commit otherwise it will run for every commit, not only when `[ -n "$TRAVIS_TAG" ]`

madig: Dammit 😁

madig:
@anthrotype anything else or can I do post3 now?

madig:
Here's hoping.

PR #96, opened 2018-07-03
Fix AppVeyor deploy, take four

madig:
This reverts commit 7456fb843a32a12ae6cfbf5c1cc8a5ab055d9dee. Try the PowerShell hack.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -68,10 +68,11 @@ artifacts:
   - path: dist\*.whl
 
 on_success:
-  - >
-    IF "%APPVEYOR_REPO_TAG%" == "true"
-    (
-    python -m pip install twine
-    &&
-    python -m twine upload dist/*.whl
-    )
+  - ps: >-
+      if ($env:APPVEYOR_REPO_TAG -eq 'true') {
+        Write-Output ("Deploying " + $env:APPVEYOR_REPO_TAG_NAME + " to PyPI...")
+        pip install --upgrade twine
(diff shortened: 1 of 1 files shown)


anthrotype:
what was wrong with the previous one?

madig:
https://ci.appveyor.com/project/rougier/freetype-py/build/1.0.99/job/8cc0lskcbi7i43fn#L1128

😢

anthrotype:
it doesn't make sense.. it does find it inside `dist` for the artifacts upload:
https://ci.appveyor.com/project/rougier/freetype-py/build/1.0.99/job/8cc0lskcbi7i43fn#L1103

jeez, Windows.. try a backslash?

anthrotype:
and this works every time with afdko repository

https://github.com/adobe-type-tools/afdko/blob/develop/.appveyor.yml#L37

madig:
> ls : Cannot find path 'C:\projects\freetype-py\tests\dist' because it does not exist.

hm?!

anthrotype:
wtf

anthrotype:
it changed directory!

anthrotype:
it should be one level up...

madig:
Ah, we `cd tests` further up... let's see.

anthrotype:
why?!

anthrotype:
ah we do that to make sure we run the tests against the installed package, not the one in the current working dir...

PR #100, opened 2018-08-22
Correct the license classifier

bgermann:
The project is BSD licensed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -123,7 +123,7 @@ setup(
         'Environment :: X11 Applications',
         'Environment :: MacOS X',
         'Intended Audience :: Developers',
-        'License :: OSI Approved :: GNU General Public License (GPL)',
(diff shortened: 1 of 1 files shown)


HinTak:
Hmm, that line came from the initial import, and in the same file at that particular commit, at the top of the file it says BSD, and it is also BSD almost everywhere else (both historically and across the source tree). So I think Nicolas really meant BSD.

BTW, only Nicolas is supposed to declare "the project is X licensed"...

rougier:
Yes, I meant BSD.

topazus:
@rougier @HinTak Hello, What is the SPDX License name of the project?

bgermann:
The license identifier is BSD-3-Clause, even though the documented license text differs in clause 3: "Neither the name of the copyright holder..." vs. "Neither the name of the freetype-py Development Team..."

PR #103, opened 2018-12-01
Add support to load face from memory file-like object

### LINKED ISSUE #32: Possible to create a Font object from memory?

yifeikong:
close #32

### DIFF
--- a/freetype/__init__.py
+++ b/freetype/__init__.py
@@ -961,12 +961,12 @@ class Face( object ):
     FreeType root face class structure. A face object models a typeface in a
     font file.
     '''
-    def __init__( self, filename, index = 0 ):
+    def __init__( self, path_or_file, index = 0 ):
         '''
         Build a new Face
 
-        :param str filename:
-            A path to the font file.
+        :param str path_or_file:
+            A path to the font file or a file-like object.
 
         :param int index:
                The index of the face within the font.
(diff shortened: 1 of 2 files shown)


freetype/__init__.py
```diff
@@ -977,17 +977,25 @@ def __init__( self, filename, index = 0 ):
...
-            error = FT_New_Face( library, u_filename, index, byref(face) )
-        except UnicodeError:
-            with open(filename, mode='rb') as f:
-                filebody = f.read()
+        if hasattr(path_or_file, 'read'):
+            filebody = path_or_file.read()
             error = FT_New_Memory_Face( library, filebody, len(filebody),
                                         index, byref(face) )
             self._filebodys.append(filebody)  # prevent gc
+            self._filename = ":memory:"
```
adrientetar: Could probably just make it None here

yifeikong: sqlite3 uses `:memory:` to denote a memory file, I thought this make sense.

madig: Using `None`, you could in code later use `if not x.filename: save or something`. Although I'm not a fan of assigning `None` to anything other than using if for default arguments, so assigning an empty string `""` would be even nicer. Unless the code at some point wants to switch to using pathlib.Path objects for all path related things.

freetype/__init__.py
```diff
@@ -977,17 +977,25 @@ def __init__( self, filename, index = 0 ):
...
+        else:
+            filename = path_or_file
+            try:
+                u_filename = c_char_p(_encode_filename(filename))
+                error = FT_New_Face( library, u_filename, index, byref(face) )
+            except UnicodeError:
```
madig: Hm. Would it make sense to just always forego FT_New_Face and use FT_New_Memory_Face instead and let Python handle the opening? One code path less.

PR #104, opened 2018-12-02
Allows `Face` object to init and attach from contents

netheril96:
So that we can load a `Face` object from anywhere, like the network or in memory computation, rather than only from disk.

### DIFF
--- a/freetype/__init__.py
+++ b/freetype/__init__.py
@@ -961,7 +961,7 @@ class Face( object ):
     FreeType root face class structure. A face object models a typeface in a
     font file.
     '''
-    def __init__( self, filename, index = 0 ):
+    def __init__( self, filename, index = 0, filebody=None ):
         '''
         Build a new Face
 
@@ -969,23 +969,33 @@ class Face( object ):
             A path to the font file.
 
         :param int index:
-               The index of the face within the font.
-               The first face has index 0.
+            The index of the face within the font.
+            The first face has index 0.
+        :param bytes filebody:
+            Contents of font file. `filename` and `filebody`
+            cannot be both set at the same time.
         '''
+        if filename and filebody:
+            raise ValueError('Both filename and filebody set')
         library = get_handle( )
         face = FT_Face( )
         self._FT_Face = None
         #error = FT_New_Face( library, filename, 0, byref(face) )
         self._filebodys = []
-        try:
-            u_filename = c_char_p(_encode_filename(filename))
-            error = FT_New_Face( library, u_filename, index, byref(face) )
-        except UnicodeError:
-            with open(filename, mode='rb') as f:
-                filebody = f.read()
+        if filebody:
+            self._filebodys.append(filebody)
             error = FT_New_Memory_Face( library, filebody, len(filebody),
                                         index, byref(face) )
-            self._filebodys.append(filebody)  # prevent gc
+        else:
+            try:
+                u_filename = c_char_p(_encode_filename(filename))
+                error = FT_New_Face( library, u_filename, index, byref(face) )
+            except UnicodeError:
+                with open(filename, mode='rb') as f:
+                    filebody = f.read()
+                error = FT_New_Memory_Face( library, filebody, len(filebody),
+                                            index, byref(face) )
+                self._filebodys.append(filebody)  # prevent gc
         if error: raise FT_Exception( error )
         self._filename = filename
         self._index = index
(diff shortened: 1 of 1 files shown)


HinTak:
I don't think it is necessary or advisable to add a new argument. It should be possible to auto-detect - if the argument to the contractor is a string, treat it as an input file name; else check that it is the right type (byte array? It is not clear what the type of filebody is meant to be, if it is not None), before treating it as a memory content; if it is neither a string nor a memory buffer, through an ArgumentException.

madig:
https://github.com/rougier/freetype-py/pull/103 tries to do the same. I suggest you keep to one argument and instead probe it to determine if it's a os.PathLike or a buffer object. Ideally, this would be done in a `@classmethod` instead of `__init__` but oh well, the API exists now.

Trying to come up with an example to probe the argument makes me realize why there are two PRs that add an extra argument. This is ugly.
```Python
class Face(object):
    def __init__(self, filename, index=0):
        library = get_handle()
        face = FT_Face()
        self._FT_Face = None
        # error = FT_New_Face( library, filename, 0, byref(face) )
        self._filebodys = []
        if hasattr(filename, "read"):
            filebody = f.read()
            error = FT_New_Memory_Face(
                library, filebody, len(filebody), index, byref(face)
            )
            self._filebodys.append(filebody)  # prevent gc
        else:
            try:
                u_filename = c_char_p(_encode_filename(filename))
                error = FT_New_Face(library, u_filename, index, byref(face))
            except UnicodeError:
                with open(filename, mode="rb") as f:
                    filebody = f.read()
                error = FT_New_Memory_Face(
                    library, filebody, len(filebody), index, byref(face)
                )
                self._filebodys.append(filebody)  # prevent gc
        if error:
            raise FT_Exception(error)
        self._filename = filename
        self._index = index
        self._FT_Face = face
```
Plus something analogous for `attach_file`. Eww.

HinTak:
Also not really sure why there are 3 different people offering 3 different solutions to the same problem - and don't seem to look at each others' approaches either...

madig:
If it's a string, freetype is going to error out and you get an exception, so that's ok.

I think the actual problem is implementing this cleanly. This is the kind of problem you get when you have code in `__init__` instead of using `@classmethod`s to construct the object...

netheril96:
> If it's a string, freetype is going to error out and you get an exception, so that's ok.

But string is a valid type for filebody. Without a separate argument, the constructor cannot distinguish between a string meant as filename and a string meant as filebody.

Anyway, I don't insist that my solution be adopted, so long as _some_ solution is merged into the mainline.

madig:
Oh, that you mean. I'd not worry too much, because FreeType is going to error out if the string doesn't point to a valid file anyway. It should be enough to distinguish between a path (basically a str in Python 2) and a buffer (has a `read` attribute).

I can see if I can come up with something that we can agree on...

anthrotype:
> It should be enough to distinguish between a path (basically a str in Python 2) and a buffer (has a read attribute).

you mean a file-like object such as an in-memory `io.BytesIO`. Python's [buffer protocol](https://docs.python.org/3/c-api/buffer.html) is something else. If you also wish to support the latter, a `frombuffer` (or `frombytes`) `@classmethod` alternative constructor is the only viable option.

To distinguish between a path (or PathLike) string and a file object, it's enough to check if the object has a `read` method, like @madig suggested. It might not be pretty, but it's a common idiom in Python. E.g. see [ElementTree.parse](https://docs.python.org/3.7/library/xml.etree.elementtree.html#xml.etree.ElementTree.parse) function which accepts either a filename or a file object as its *source* parameter.

netheril96:
It just feels so hacky to me if we rely on the fact that filenames cannot be valid font contents. What if there are very weird filenames?

yifeikong:
@HinTak Actually, I'm using freetype with a very small font file which only contains a dozen chars. The purpose of this font is to encrypt or obfuscate the content, I use freetype to get the read char. So IMHO filesize should not be used to determine what the object means. Besides, duck typing is a common idiom in Python, if it has a `read` method, it must be a file-like object.

netheril96:
> For python 3, string is different from a memory buffer; the distinction is less clear for python 2, I admit. 

In Python 3, filename can be both bytes and Unicode string. This is needed on Unix like systems, because in POSIX, filename can be any sequence of bytes excluding '\0', notwithstanding the encoding settings.

rougier:
@madig Ouch, sorry for that.

madig:
Support in master.

PR #106, opened 2018-12-11
Implement Face loading from memory

madig:
The fourth PR to tackle loading a font from memory instead of from disk. 

- Since we still have to support Python 2(?), testing the `file_or_stream` argument for being a `bytes` object won't work, accept a `io.BytesIO` object instead.
- Implement @rougier's suggestion at https://github.com/rougier/freetype-py/pull/33#issuecomment-167834852. Since loading from a path is the default, I did not add a `from_file` class method. It would make more sense when `Face::__init__` is refactored to not have any logic and just stores what it's given.

Not sure what to do about `Face::_filename`. it doesn't seem to be used anywhere.

Working on `attach_file`, I need a test case.

Closes https://github.com/rougier/freetype-py/pull/33, https://github.com/rougier/freetype-py/pull/103, https://github.com/rougier/freetype-py/pull/104

### DIFF
--- a/freetype/__init__.py
+++ b/freetype/__init__.py
@@ -15,6 +15,7 @@ Note: C Library will be searched using the ctypes.util.find_library. However,
       this search might fail. In such a case (or for other reasons), you may
       have to specify an explicit path below.
 '''
+import io
 import sys
 from ctypes import *
 
@@ -961,12 +962,12 @@ class Face( object ):
     FreeType root face class structure. A face object models a typeface in a
     font file.
     '''
-    def __init__( self, filename, index = 0 ):
+    def __init__( self, path_or_stream, index = 0 ):
         '''
         Build a new Face
 
-        :param str filename:
-            A path to the font file.
+        :param Union[str, io.BytesIO] path_or_stream:
+            A path to the font file or an io.BytesIO stream.
 
         :param int index:
                The index of the face within the font.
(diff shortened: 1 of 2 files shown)


freetype/__init__.py
```diff
@@ -961,12 +962,12 @@ class Face( object ):
...
     font file.
     '''
-    def __init__( self, filename, index = 0 ):
+    def __init__( self, path_or_stream, index = 0 ):
         '''
         Build a new Face
 
-        :param str filename:
-            A path to the font file.
+        :param Union[str, io.BytesIO] path_or_stream:
```
anthrotype: Why just BytesIO? One could well wish to pass an already open binary file object.
You could use typing.BinaryIO to catch both

freetype/__init__.py
```diff
@@ -961,12 +962,12 @@ class Face( object ):
     FreeType root face class structure. A face object models a typeface in a
     font file.
     '''
-    def __init__( self, filename, index = 0 ):
+    def __init__( self, path_or_stream, index = 0 ):
```
anthrotype: path_or_file would also be ok.

freetype/__init__.py
```diff
@@ -975,22 +976,39 @@ def __init__( self, filename, index = 0 ):
...
+    def _init_from_memory(self, library, face, index, byte_stream):
+        error = FT_New_Memory_Face(
+            library, byte_stream, len(byte_stream), index, byref(face)
+        )
+        self._filebodys.append(byte_stream)  # prevent gc
+        return error
+
+    @classmethod
+    def from_memory(cls, bytes_, index=0):
+         return cls(io.BytesIO(bytes_), index)
```
anthrotype: Why not call _init_from_memory directly here, instead of making a temp BytesIO?

anthrotype: Also, frombytes (or if you like underscores from_bytes) reads better to me, but that’s just taste

madig:
Bah. Can't get attaching bytes to work because `FT_Open_Args` -> memory_base expects `c_ubyte` when it should be `POINTER(c_ubyte)`. Did this ever work?

freetype/__init__.py
madig: It's a `@classmethod` and I have to go through `__init__` somehow, no?

freetype/__init__.py
madig: Hm, guess it doesn't hurt in a docstring.

freetype/__init__.py
anthrotype: not necessarily. You can call `self = cls.__new__(cls)` to allocate the new object and then initialize it

anthrotype: basically, calling `__new__` explicitly in the alternative constructor would circumvent `__init__`

anthrotype: never mind, it's also fine like that. It doesn't add too much overhead to create the BytesIO (it's just a thin wrapper around some bytes string), and the whole from_bytes classmethod is mostly syntactic sugar, sparing the user an extra `from io import BytesIO` basically.

anthrotype:
from the CI failures, I think you need to update the multibuild submodule

freetype/__init__.py
madig: Hm, I'd need to replicate the init logic then, no? This is something for a refactoring with breaking changes...

freetype/__init__.py
madig: Had that before, but technically I expect a stream here...

madig:
I lost patience with `attach_file`, no idea what the problem is. Maybe in another PR.

madig:
@rougier Please have a look. Can merge and tag if you give your ok.

madig:
https://github.com/rougier/freetype-py/pull/106/files#diff-c981959bb975a2ca7046d119aeb25b02L990

I removed it in this PR, but maybe you know what it's used for?

rougier:
I can imagine the idea was to keep the reference to the filename but I'm not sure it is actually used elsewhere.

madig:
I couldn't find a reference to it in the code 🤔 I can leave it in if it's unclear, but only assign when I think a path is passed in?

rougier:
Well, if it is not used, maybe it's better to remove it completely.

madig:
I like your attitude. So! Anything else to do here?

rougier:
Nope. Let's pull the trigger.

PR #108, opened 2019-03-17
FreeType 2.10, HarfBuzz 2.3.1: Update libs, add Python 3.7 to build list

madig:


madig:
D'OH. I forgot that I don't need to add Python 3.7 because we don't depend on CPython internals.

PR #110, opened 2019-03-17
Remove Python 3.7 from CI build

madig:
The packages for 3.6 are universal because we don't depend on CPython internals. Duh.


anthrotype:
I would have removed 3.6 instead ...

madig:
Mh, next time.

PR #118, opened 2020-01-10
Improve Variable Font support

josh-hadley:
Updates to several freetype-py modules to improve support for Variable Fonts. Includes tests for added code, as well as an example demonstrating variable font usage.


freetype/__init__.py
```diff
@@ -18,13 +18,22 @@
 import io
 import sys
 from ctypes import *
```
rougier: No connected to the PR but at some point we should get rid of this `from ctypes import *` in favor of `import ctypes`

rougier:
Really nice ! Thanks for the PR. This looks good to me. Let's wait a day or two for potential comments and then we merge (maybe you'll need to remind me).

HinTak:
FWIW, "from ctype import *" and "import ctypes" do different things. In the latter, you still need to do "ctypes.x" in code, vs bare "x" in the former.

freetype/__init__.py
josh-hadley: Agree. There are lots of other code style/PEP8 & related issues that would be good to address (and maybe enforce, via CI) eventually. I didn't want to pollute this PR with such updates though 😄

josh-hadley:
> FWIW, "from ctype import *" and "import ctypes" do different things. In the latter, you still need to do "ctypes.x" in code, vs bare "x" in the former.

Right. My preference would be `from ctypes import (<sequence of each individual thing used from ctypes>)`, so that only the import statement(s) would change, rather than each line of code where a ctypes item is used. We can discuss more in a later PR 😄

rougier:
@HinTak Yes but I (personally) prefer to prefix, maybe because I'm using numpy a lot and, from time to time, I also need the math module with a lot of overlap/conflicts. Anyway, I can live with the `import *` if everyone's ok.

HinTak:
I like what @josh-hadley says, an explicit "from ctypes import x, y, z" ... as long as the list does not get too long. But no need to change what's currently working...

Not a big fan of numpy - I think if you name something the same as a more common package's, you better be 100% compatible...

madig:
For what it's worth, there is https://pypi.org/project/autoflake/ and maybe https://github.com/quentinsf/dewildcard that can automatically expand star imports.

madig:
Woops! You might want to run `pylint --disable=all --enable=E Lib` to find more typos and stuff.

josh-hadley:
> Woops! You might want to run `pylint --disable=all --enable=E Lib` to find more typos and stuff.

Yeah, I have been using flake8, but had disabled a bunch of warnings to keep the noise down from issues in the existing code...this would've been caught by flake8. It would be good to have pylint or flake8 or similar running in CI to catch these on PR submit :smile:

Related: the appveyor CI failure does not appear to be caused by the most recent commit (https://github.com/rougier/freetype-py/commit/83d0c53255). Seems like a timeout attempting to retrieve build components.

madig:
Hrm, yeah, we download FreeType and HarfBuzz release tarballs so we can build a static Python wheel. Not sure what's going on there. Needs to be updated anyway, in a separate PR...

Agree on a clean-up PR with linting and stuff.

rougier:
Merged ! Thanks again for the nice PR.

PR #122, opened 2020-05-05
Drop debian directory

### LINKED ISSUE #121: Drop debian directory
Please drop the debian directory. It is pretty outdated. I created a py3-only freetype-py Debian package that is in bullseye (current testing). It's already available on Ubuntu Focal Fossa.

bgermann:
Fixes #121.


HinTak:
Hmm, I kind of meant updating it for python3 (and current debian/ubuntu - which the reporter said had done), rather than just deleting it.

bgermann:
I am the reporter of the issue and you can find the package at https://tracker.debian.org/pkg/freetype-py

The idea of deleting the directory is that there is one version of the package and I do not have to filter the debian directory from the upstream files.

HinTak:
I meant updating the debian directory to something close enough to what debian/ubuntu are shipping or willing to ship.

PR #127, opened 2020-09-02
Fix bbox calculation in hello-world example

satirebird:
If the string contains a ',' the example code produces the following error:
```
  File "hello-world.py", line 47, in <module>
    Z[y:y+h,x:x+w] += numpy.array(bitmap.buffer, dtype='ubyte').reshape(h,w)
ValueError: operands could not be broadcast together with shapes (0,28) (35,28) (0,28)
```
If the string in the hello-world example contains some characters which passes the baseline the bbox calculation fails. This will be the case when a ',' is added to the example string.

### DIFF
--- a/examples/hello-world.py
+++ b/examples/hello-world.py
@@ -13,14 +13,15 @@ if __name__ == '__main__':
     import matplotlib.pyplot as plt
 
     face = Face('./Vera.ttf')
-    text = 'Hello World, !'
-    face.set_char_size( 48*64 )
+    text = 'Hello World !'
+    fontsize = 48
+    face.set_char_size( fontsize*64 )
     slot = face.glyph
 
     # First pass to compute bbox
     width = 0
-    top = -32768
-    bot = 32767
+    top = -fontsize #The distance from baseline to the top row of the bbox
+    bot = fontsize #The distance from baseline to the bottom row of the bbox (may be negative)
     previous = 0
     for i,c in enumerate(text):
         face.load_char(c)
--- a/examples/hello-world.py
+++ b/examples/hello-world.py
@@ -13,23 +13,25 @@ if __name__ == '__main__':
     import matplotlib.pyplot as plt
 
     face = Face('./Vera.ttf')
-    text = 'Hello World !'
+    text = 'Hello World, !'
     face.set_char_size( 48*64 )
     slot = face.glyph
 
     # First pass to compute bbox
-    width, height, baseline = 0, 0, 0
+    width = 0
+    top = -32768
+    bot = 32767
     previous = 0
     for i,c in enumerate(text):
         face.load_char(c)
         bitmap = slot.bitmap
-        height = max(height,
-                     bitmap.rows + max(0,-(slot.bitmap_top-bitmap.rows)))
-        baseline = max(baseline, max(0,-(slot.bitmap_top-bitmap.rows)))
+        top = max(top, slot.bitmap_top)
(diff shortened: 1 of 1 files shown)


HinTak:
On the whole, I don't quite agree with this kind of enhancement: examples are meant to be simple and illustrative. Bolting on arbitrary features like supporting descenders - and artificially inserting a "," from "Hello World!" to "Hello World, !" , seems daft.

A better approach might be to actually create a **new** example, with a valid phrase like "Surprise!" (where the "p" goes below the baseline), plus something **else** besides just coping with descenders.

satirebird:
The main reason for the fix is that the original code calculates a baseline and height with some none intuitive computations. 
```
height = max(height, bitmap.rows + max(0,-(slot.bitmap_top-bitmap.rows)))
baseline = max(baseline, max(0,-(slot.bitmap_top-bitmap.rows)))
```
In addition it does not work in some cases. The rendering of "Hp" works while the rendering of "H," produces a negative index to the np array. I used the code snippet  in my software and I search a while for the error. I want to keep other users falling into the same issue.
I think the change make the example code more clear and robust.

Of course adding a comma to the "Hello World !" string is unnecessary. It should be removed.

HinTak (changes requested):
I think the idea of the second part of the change - making the calculation cope with characters with descenders - is fine, but the first part contains too many questionable items: the arbitrary and extra "," and the wrong / arbitrary numbers 3276x. 

Also, I haven't actually checked the calculation - the idea is always that you do one pass to find the extent, a second pass to do the actual drawing. I know `baseline` shouldn't be called that so that's fine, but you removed the `left` variable too... That reminds me - if you want this code to work correctly for arbitrary strings, you need to make one more set of adjustments: besides coping with descenders, you also need to take into account of the left-side-bearing of the first character, and the right-side-bearing of the last character. i.e when the first character draws before its origin, and the last character draws after its advance-width.

satirebird:
> I know baseline shouldn't be called that so that's fine, but you removed the left variable too...

The left variable wasn't used so I removed it.

> if you want this code to work correctly for arbitrary strings, you need to make one more set of adjustments: besides coping with descenders, you also need to take into account of the left-side-bearing of the first character, and the right-side-bearing of the last character. i.e when the first character draws before its origin, and the last character draws after its advance-width.

You are right. But the example should be as simple as possible. This code works in all cases because the first kerning is not negative in all cases (due to previous character is NULL). The example will not write outside the previously calculated bbox.

HinTak (approved):

HinTak:
Hi @satirebird , thanks for the changes. I have finally got round to try it out - it does what it is supposedly to do, but curiously the dimension of the output has changed from (38,306) to (37,306), 1 pixel shorter. It can be seen that in the new output, the top of the "l"'s are closer to the edge than previous.

It may be a genuine improvement, but I think we need to understand where that 1-pixel difference comes from.

Btw, my harfbuzz-python-demos repo have some variant of this code adapted for Arabic , which, besides having a different text-run direction, also have ascender/descender/left-side-bearing/right-side-bearing per glyph differing from the font's design bound box by quite a lot. Not meant to be bullet proof either, just rendering each of the given phrases in box...

HinTak:
I think the 1-pixel difference is buried somewhere in the change from:
```
height = max(height,
                      bitmap.rows + max(0,-(slot.bitmap_top-bitmap.rows)))
```

```
     height = top - bot
```

I don't think I care enough either way, so let's just merge this.

Issue #123, opened 2020-05-19
build embedded freetype lib with color font support?


anthrotype:
Currently the freetype2 embedded library that freetype-py includes is built without PNG support so loading emoji fonts doesn't work. One has to build their own freetype2 library with color support and have freetype-py load from that.

the setup-build-freetype.py script is configured to disable external dependencies, for good reasons since we want to make sure the compiled libarary is stabdalone and portable:
https://github.com/rougier/freetype-py/blob/c578643b5736130c4341584655cb9813c32b5482/setup-build-freetype.py#L143-L146

Maybe we could see if we can also build the other external dependencies (zlib, png, perhaps eve harfbuzz) and statically link them into the shared library that we are embedding in the freetype-py distribution.

anthrotype:
correction sorry we already support harbuzz in the setup-build-freetype.py script:
https://github.com/rougier/freetype-py/blob/c578643b5736130c4341584655cb9813c32b5482/setup-build-freetype.py#L161

rougier:
Stupid question: do we need the "official"  png library of can freetype uses something like https://github.com/nothings/stb/blob/master/stb_image.h ?

HinTak:
Freetype uses the official libpng.

PR #133, opened 2020-12-22
add support for building zlib and libpng libraries in setup-build-freetype.py

mammo0:
Hello,

this PR is for solving #123.

Building with zlib or libpng support is disabled by default. So there's no new default behavior of this script.
But if you want to build with

- zlib support, set the environment variable `FREETYPEPY_WITH_ZLIB` to some value, e.g. `1`
- libpng support, set the environment variable `FREETYPEPY_WITH_LIBPNG` to some value, e.g. `1`

### DIFF
--- a/setup-build-freetype.py
+++ b/setup-build-freetype.py
@@ -33,6 +33,16 @@ HARFBUZZ_HOST = "https://www.freedesktop.org/software/harfbuzz/release/"
 HARFBUZZ_TARBALL = "harfbuzz-2.6.7.tar.xz"
 HARFBUZZ_URL = HARFBUZZ_HOST + HARFBUZZ_TARBALL
 HARFBUZZ_SHA256 = "49e481d06cdff97bf68d99fa26bdf785331f411614485d892ea4c78eb479b218"
+BUILD_ZLIB = os.environ.get("FREETYPEPY_WITH_ZLIB", "")
+ZLIB_HOST = "https://zlib.net/"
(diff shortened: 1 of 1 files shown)


HinTak:
If you bundle very common libraries like zlib and libpng you might need to look into the symbol export systems (not sure about it under cmake etc, but it is part of autoconf'ed build) not to re-export 3rd-party symbols, to avoid the Linux version of dll hell - where software using freetype **and*/ libpng get confused by the bundled libpng symbols within your build of freetype.

mammo0:
Well, honestly I've got not the knowledge of how the re-export of the symbols can be prevented.

I have oriented myself on how `harfbuzz` was built into the freetype library. After it worked for me, I just opened the PR here...

HinTak:
The autoconf build of freetype sets symbols to be by default hidden, and passes a linker script to the linker to explicitly declare which symbols are exported.

mammo0:
Ok, can you show me an example?

HinTak:
You can each for export in freetype e.g. https://git.savannah.gnu.org/cgit/freetype/freetype2.git/commit/?id=771d3c8ae25d5042a3f6b76a329ad3dd782eca66

mammo0:
So the variable `EXPORTS_LIST` should be set?

I'm sorry, but I don't have much experience in C/C++ and library programming... So maybe you can make the necessary changes? Because it seems you know what you're talking about...

PR #134, opened 2020-12-23
support for ARM builds

mammo0:
Hello,

during the build process on a Raspberry Pi I got an error from cmake, that the `-m32` or `-m64` flag is not supported on this platform.

So I reworked the code to detect not only if the system is 32/64bit but also if it's an ARM system. There's also support for `aarch64`.

### DIFF
--- a/setup-build-freetype.py
+++ b/setup-build-freetype.py
@@ -81,10 +81,7 @@ if sys.platform == "darwin":
 if "linux" in sys.platform:
     c_flags = cxx_flags = "-O2"
     ld_flags = ""
-    if (
-        platform.machine() == "x86_64"
-        or platform.machine() == "x86"
-    ):
+    if platform.machine() == "x86_64":
         if (
             os.environ.get("PYTHON_ARCH", "") == "32"
             or platform.architecture() == "32bit"
--- a/setup-build-freetype.py
+++ b/setup-build-freetype.py
@@ -21,6 +21,7 @@ import tarfile
 import urllib
 from os import path
 from urllib.request import urlopen
+import platform
 
 # Needed for the GitHub Actions macOS CI runner, which appears to come without CAs.
 import certifi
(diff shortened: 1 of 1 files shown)


HinTak:
You are going about it the wrong way - multi-arch build is only ever useful when the cpu is x86_64 (Linux /windows /mac os...) so you should conditioned on `if x86_64`. I heard that there are multi-arch raspberry pi distro/systems out there with armhf+aarch64 too though - not sure if they are switchable by -m32/m64.

Anyway, why do you want to build freetype on pi when it is available system-wide (and "./configure && make" works?)

HinTak:
I'd suggest you change it to conditioning on x86_64 / x64 (for windows too) instead ; but mixed-arch raspberrypi system exists (64-bit kernel with 32-bit userland, etc), though not yet common, so I am not sure what purpose this change serve.

mammo0:
First thanks for the quick response. But I don't really get your point...

My purpose is not a multiarch build on a x86_64 machine. I had just the issue, that if I manually build the freetype library with `FREETYPEPY_BUNDLE_FT` on my Pi, cmake complained about the `-m32`/`-m64` flag. And it seems that those flags are not supported on arm. So I added a new case, that checks if the build system is on running on ARM. If so, it simply doesn't add those flags.

> Anyway, why do you want to build freetype on pi when it is available system-wide (and "./configure && make" works?)

I stumbled about this issue because for my project I needed a freetype library, that was compiled with `libpng` support (see my PR #133). After successfully building the library with `libpng` on my x86_64 PC, I tried it on my Pi. There I got the above mentioned error.

HinTak:
You tested on `platform != aarch64`. But aarch64 is not the one that needs special treatment. x86_64 / x64 is.

mammo0:
I test on `platform != arm` and `platform != aarch64` to check if it's ARM or not... Otherwise the the old behavior is used.

HinTak:
You should test on `platform == x86_64` instead. I wrote at the very beginning. That said, biarch armhf/aarch64 systems, although much rarer than x86_64 biarch systems,  are around.

mammo0:
Ok, then I will switch the if-statement.

mammo0:
Hello @HinTak,
please check my latest commit (f0c6bd8) in this PR. Is that how you meant it?

setup-build-freetype.py
```diff
@@ -81,8 +81,10 @@
 if "linux" in sys.platform:
     c_flags = cxx_flags = "-O2"
     ld_flags = ""
-    if ("arm" not in platform.machine() and 
-            platform.machine() != "aarch64"):
+    if (
+        platform.machine == "x86_64"
+        or platform.machine == "x86"
+    ):
```
HinTak, line 87: No need for "x86"...

mammo0: Ok can change that. I added the second case if the script is executed on an 32bit machine... Don't know if the output of `machine` is there also `x86_64`.

By the way, I missed also the `()` after the `machine` function call.

HinTak:
The change, as is, is okay now, I think, but I just spotted a "problem" in the surrounding code: as I said, the special case is x86_64, not linux. People routinely use both 32-bit and 64-bit freetype on 64-bit (intel) windows too, and both 32-bit and 64-bit freetype on 64-bit Mac OS too, until OS X 15 (when they officially dropped 32-bit support, I think). So special-case'ing on linux seems wrong.

Anyway, I would say merge this as is now. If somebody care enough about bi-arch on 64-bit windows and mac os x, they can add some more code later.

HinTak:
gcc on Mingw and Mac OS X takes -m32/-m64  too, like their linux siblings.

mammo0:
> Anyway, I would say merge this as is now. If somebody care enough about bi-arch on 64-bit windows and mac os x, they can add some more code later.

Thank you for approving this.

PR #139, opened 2021-11-22
Add arm64 to build matrix and update cibuildwheel

### LINKED ISSUE #138: M1 compatibility
I noticed you are using cibuildwheel; it should be relatively easy to provide wheels for apple silicon. Are there any special reasons you are not doing so today?

Korijn:
Closes #138

* This upgrades CI to use the latest version of cibuildwheel and adds apple silicon wheels to the build configuration.
* Apple silicon wheels can only be built for python >=3.8.
* This newer version of cibuildwheel also happens to build musllinux wheels, so that's a free bonus, I guess!
* Split out the publish to pypi job

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,54 +13,66 @@ on:
 
 jobs:
   build:
-    name: Build wheels on ${{ matrix.os }}
+    name: Build wheels on ${{ matrix.os }} ${{ matrix.name }}
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
-
+        include:
+        - os: ubuntu-latest
+          cibw_build: "cp36-*"
+        - os: windows-latest
+          cibw_build: "cp36-*"
+        - os: macos-latest
+          cibw_build: "cp36-*"
+        - os: macos-latest
+          cibw_build: "cp38-macosx_arm64"
+          name: '(arm64)'
+        - os: macos-latest
+          cibw_build: "cp38-macosx_universal2"
+          name: '(universal2)'
     steps:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 0  # unshallow fetch for setuptools-scm
 
-    - name: Use MSBuild (Windows)
-      uses: microsoft/setup-msbuild@v1.0.2
-      if: matrix.os == 'windows-latest'
-
-    - uses: actions/setup-python@v1
-      name: Install Python
-      with:
-        python-version: '3.6'
-
-    - name: Install cibuildwheel
-      run: |
-        python -m pip install cibuildwheel==1.5.2
-
-    - name: Build wheel
-      run: |
-        python -m cibuildwheel --output-dir dist
+    - name: Build wheels
+      uses: pypa/cibuildwheel@v2.2.2
       env:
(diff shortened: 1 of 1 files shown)


rougier:
Waiting for build before merging.

almarklein:
Side question: to what extend can we expect FreeType to be available on Linux systems? If it's always there, then we don't have to bother with the Linux-aarch64 builds (and others).

Korijn:
> Side question: to what extend can we expect FreeType to be available on Linux systems? If it's always there, then we don't have to bother with the Linux-aarch64 builds (and others).

As far as I know freetype is not listed in any of the "manylinux" policies, so at least _they_ don't seem to think it's widely available.

Korijn:
I pushed two small changes that make the builds a tiny bit faster.

Korijn:
I think this can be merged!

.github/workflows/ci.yml
```diff
@@ -25,36 +25,32 @@ jobs:
...
-    - name: Build wheel
-      run: |
-        python -m cibuildwheel --output-dir dist
+    - name: Build wheels
+      uses: pypa/cibuildwheel@v2.2.2
       env:
         # The packaged FreeType library is independent of the Python ABI so we only
         # need to build it once.
-        CIBW_BUILD: "cp36-*x86_64 cp36-*amd64"
+        CIBW_BUILD: "cp39-*"
```
anthrotype, line 33: the reason it was built on a lower python minor version was to make sure that the tests pass on the minimum supported version. Arguably 3.6 is too low nowadays (EOL soon at end of year), how about setting it to 3.7?

.github/workflows/ci.yml
```diff
@@ -25,36 +25,32 @@ jobs:
...
+        output-dir: dist
 
     - uses: actions/upload-artifact@v2
       with:
         name: freetype-py-dist
         path: dist/*.whl
 
+    - uses: actions/setup-python@v1
+      name: Install Python
+      if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
```
anthrotype, line 50: the whole building of sdist and publishing of wheels could be moved to a separate "deploy" job that is conditioned on the "build" job passing and is only triggered on tagged commits (so you don't have to repeat the `if:` condition every time).

anthrotype (approved):
LGTM with minor comments

.github/workflows/ci.yml
Korijn: I set it back to 3.6 or 3.7, whatever you like, but arm64 needs 3.8 at a minimum, macos arm64 wheels aren't supported below that.

.github/workflows/ci.yml
Korijn: Sure, I also prefer that but I didn't want to refactor your CI pipeline without consent :) I guess this is the green light to do that then.

.github/workflows/ci.yml
Korijn: Done.

.github/workflows/ci.yml
Korijn: Done.

Korijn:
Ready for another review round.

rougier:
Waiting for output before merging (I will need some ping or I'll forget to merge...)

Korijn:
Passed!

Korijn:
@rougier I think you can merge now :)

Korijn:
One more ping @rougier to get this merged! :)

rougier:
Merged!

PR #141, opened 2021-11-24
Add support for SDF rendering.

esoma:
- Ups the freetype version to 2.11
- Adds the `FT_RENDER_MODE_SDF` flag
- Adds the `render` method to `GlyphSlot` corresponding to `FT_Render_Glyph`

There is no `FT_LOAD_TARGET_XXX` for SDF, so an api for `FT_Render_Glyph` is required to render an SDF. Not sure if this is an oversight in the freetype API or by design.

### DIFF
--- a/freetype/__init__.py
+++ b/freetype/__init__.py
@@ -975,6 +975,98 @@ class GlyphSlot( object ):
           glyph: valid FT_GlyphSlot object
         '''
         self._FT_GlyphSlot = slot
+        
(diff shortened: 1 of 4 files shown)


rougier:
Oh wait, you mean freetype has now a native SDF render mode? Since which version?

esoma:
2.11

rougier:
Fantastic!

PR #143, opened 2022-01-24
Expose binding to FT_Outline_EmboldenXY()

takaakifuji:
I noticed that [`FT_Outline_EmboldenXY()`](https://freetype.org/freetype2/docs/reference/ft2-outline_processing.html#ft_outline_emboldenxy) looks missing in `raw.py`, which is an enhanced variant to `FT_Outline_Embolden()`.

Adding a Pythonic binding should be trivial, but I'm not sure it's worth it.

### DIFF
--- a/freetype/raw.py
+++ b/freetype/raw.py
@@ -269,6 +269,7 @@ FT_Outline_Copy                = _lib.FT_Outline_Copy
 FT_Outline_Decompose           = _lib.FT_Outline_Decompose
 FT_Outline_Done                = _lib.FT_Outline_Done
 FT_Outline_Embolden            = _lib.FT_Outline_Embolden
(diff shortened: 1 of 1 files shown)


HinTak:
Please add "try ... except AttributeError:" like other places in raw.py, and perhaps also a comment about what version of freetype the routine was introduced. This construct is there to allow version differences and for adding recently introduced routines.

```
try:

except AttributeError:
    pass
```

takaakifuji:
Thank you so much for pointing this out! I didn't realise the function was relatively new one.

HinTak:
Apparently it first appeared in 2.5 (released in 2013), so not particularly new, but certainly rarely used.

rougier:
Thanks.

PR #144, opened 2022-01-31
Add aarch64 builds to CI build matrix

Korijn:
This adds aarch64 builds for manylinux and musllinux using QEMU. That makes the wheel installable on e.g. raspberry pi.

I added them as separate jobs to the build matrix as they are quite slow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,29 +21,59 @@ jobs:
         include:
         - os: ubuntu-latest
           cibw_build: "cp36-*"
+          cibw_archs_linux: "auto64"
+          cibw_archs_macos: "auto64"
         - os: windows-latest
           cibw_build: "cp36-*"
+          cibw_archs_linux: "auto64"
+          cibw_archs_macos: "auto64"
         - os: macos-latest
           cibw_build: "cp36-*"
+          cibw_archs_linux: "auto64"
+          cibw_archs_macos: "auto64"
         - os: macos-latest
           cibw_build: "cp38-macosx_arm64"
+          cibw_archs_linux: "auto64"
+          cibw_archs_macos: "arm64"
           name: '(arm64)'
         - os: macos-latest
           cibw_build: "cp38-macosx_universal2"
+          cibw_archs_linux: "auto64"
+          cibw_archs_macos: "universal2"
           name: '(universal2)'
+        - os: 'ubuntu-latest'
+          cibw_build: "cp36-manylinux*"
+          cibw_archs_linux: 'aarch64'
+          cibw_archs_macos: "auto64"
+          name: '(manylinux aarch64)'
+          qemu: true
+        - os: 'ubuntu-latest'
+          cibw_build: "cp36-musllinux*"
+          cibw_archs_linux: 'aarch64'
+          cibw_archs_macos: "auto64"
+          name: '(musllinux aarch64)'
+          qemu: true
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -21,29 +21,59 @@ jobs:
         include:
         - os: ubuntu-latest
           cibw_build: "cp36-*"
+          cibw_archs_linux: "auto64"
+          cibw_archs_macos: "auto64"
```
Korijn, line 25: If anyone knows how to avoid this ridiculous level of duplication, that would be much appreciated.

djhoese: I *think* you could define these under `matrix` and then override the default for the aarch64 element. Depending on how much work you want to do, you may also be able to combine all this logic in the environment variables for cibuildwheel: https://github.com/vispy/vispy/blob/182f87d62baa7158db96d963c7da435b723178cc/.github/workflows/wheels.yml#L22-L35

Korijn: > I think you don't need to set `cibw_archs_macos` matrix variable when running on ubuntu, nor `cibw_archs_linux` or `cibw_archs_macos` when running on windows, etc. They will default to empty variable and should be ignored anyway because they are not used by cibuildwheel unless on the target platform.

See here what happened when I did that: https://github.com/rougier/freetype-py/runs/5008175186?check_suite_focus=true#step:4:45

`ValueError: '' is not a valid Architecture`

Korijn: > I _think_ you could define these under `matrix` and then override the default for the aarch64 element.

Yeah so _how exactly_ that can be done is unclear to me.

> Depending on how much work you want to do, you may also be able to combine all this logic in the environment variables for cibuildwheel: https://github.com/vispy/vispy/blob/182f87d62baa7158db96d963c7da435b723178cc/.github/workflows/wheels.yml#L22-L35

As you can see from the currently running build, the aarch64 jobs independently take _very_ long because they run via QEMU. Using one job per OS would go over global build timeouts, and be a pretty terrible developer experience, I'm pretty sure.

djhoese: > Using one job per OS would go over global build timeouts

Isn't that what's already happening? One matrix "entry" for each operating system/arch? I'm also not sure I understand the `cibw_build` in this matrix. Some of these look like they are only building `cp36-*` wheels? Or only 3.8?

I might not know what I'm talking about. It's been a while since I've tried to get cibuildwheel to do what I want.

rougier:
Jumping late in the discussion. Is it working properly now?

Korijn:
> Jumping late in the discussion. Is it working properly now?

Nah I messed up, will get back to it tomorrow

Korijn:
I think I got it now, squashed my commits to erase all the trial-and-error history (no value in keeping that around).

This time it should build in about 20 minutes, instead of 2 and a half hours. :)

rougier:
Many thanks!

Korijn:
Just as an FYI this is the list of wheels that are now being built on CI:

image

djhoese:
So now freetype-py just needs another release to get those wheels on PyPI, right? Thanks @Korijn!

Korijn:
> So now freetype-py just needs another release to get those wheels on PyPI, right? Thanks @Korijn!

That's right.

rougier:
There's a pending PR that shoudl be merged soon and then we can do a release.

PR #147, opened 2022-03-09
use print as function instead of statement (Python 3)

basnijholt:


Korijn:
@rougier I think this can be merged?

rougier:
Yes, sorry for long delay.

PR #152, opened 2022-05-05
Remove release notes in docs

almarklein:
Since the release notes are now on the GH release pages.


almarklein:
Also, the latest build on RTD is from 3 years ago :)   https://readthedocs.org/projects/freetype-py/builds/
Anything I can do to help with that?

rougier:
Yes, I need to upload the new doc. Best would be to have a github action to do that automatically. Any experience on that?

almarklein:
Normally RTD will just rerender automatically on new commits to main, but I guess the webhook is broken.

almarklein:
If you're logged into readthedocs.org, go to the project, and then the admin panel. Then it's under "integrations"
image

rougier:
I configured everything on readthedocs but the doc still seems outdated.

HinTak:
It looks like the current ci runs twine manually, but even twine itself doesn't do that any more:

https://github.com/pypa/twine/blob/fa8e3d747959756ad20b3a26ef8a1fcb4f9cdc93/.github/workflows/release.yml#L62

HinTak:
I believe this is relevant: https://github.com/pypa/twine/issues/999 . FWIW, skip-python moved to trusted publisher pypi workflow in January, because the `secrets.PYPI_USERNAME` and `secrets.PYPI_PASSWORD` stopped working, I think.

HinTak:
We definitely don't have a `.readthedocs.yaml` so that's that.

HinTak:
Okay, kind of unrelated, but the docs stopped updating between Dec 17, 2018 and May 15, 2022. Then it broke in aug 2024 between 23rd and 21st.

Anyway, it works now.

While I think something needs to be done about the massively outdated release notes, I rather think importing them from the release page to the doc directory is better than deleting them - to keep the docs in one place.

HinTak:
Keeping the release notes inside the repo is also perhaps more appropriate in light of the possibility that the repo may move - and github based release notes will be lost in such cases (#188).

PR #162, opened 2023-02-20
Add pyinstaller hook

Korijn:
This adds a pyinstaller hook to freetype which is made discoverable using an entry point. That means that if projects wish to freeze their python application using pyinstaller, pyinstaller will automatically know that it should include the freetype binary in the package.

It includes a pytest setup that is also run on CI, which ensures that freetype can be imported from a pyinstaller-frozen application. This makes sure the hook won't be broken in the future.

(Reason for this change is that we are ensuring wgpu-py and pygfx are pyinstaller compatible by default, and freetype became one of our dependencies, so I'm contributing pyinstaller support coming from that angle :)


rougier:
Thanks. I'm not too familiar with pyinstaller but your changes seem reasonable.

PR #167, opened 2023-05-06
Add missing build-backend to pyproject.toml

mgorny:
Add missing `build-backend` declaration to `pyproject.toml`. The existence of `requires` key without a specific backend triggers an error in Gentoo packaging, as it indicates an incomplete port to PEP 517.


mgorny:
Thanks!

PR #168, opened 2023-05-06
Remove redundant wheel dep from pyproject.toml

mgorny:
Remove the redundant `wheel` dependency, as it is added by the backend automatically.  Listing it explicitly in the documentation was a historical mistake and has been fixed since, see: https://github.com/pypa/setuptools/commit/f7d30a9529378cf69054b5176249e5457aaf640a

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools>=42", "wheel", "setuptools_scm[toml]>=3.4", "certifi", "cmake"]
+requires = ["setuptools>=42", "setuptools_scm[toml]>=3.4", "certifi", "cmake"]
 build-backend = "setuptools.build_meta"


rougier:
Seems the build failed.

anthrotype:
I think this PR needs to be rebased on master because it also needs this https://github.com/rougier/freetype-py/pull/167

mgorny:
Honestly, this failure makes no sense. Lemme try locally, perhaps cibuildwheel is broken (sigh).

mgorny:
Ok, now I see what the problem is. If `FREETYPEPY_BUNDLE_FT` is set, `setup.py` imports `wheel` directly. This also happens when `setup.py` is run to get build-time dependencies, so we're dealing with a circular dep — `setup.py` needs `wheel` to be able to say that it needs `wheel`.

The `wheel` dep is incorrect for the general case but needed when `FREETYPEPY_BUNDLE_FT` is set.

PR #170, opened 2023-05-12
don't depend on cmake via pip

landfillbaby:
should just rely on a system copy of cmake. fixes build on termux.

typo in the commit message oops


HinTak:
You can force-push to fix typos. (Better still fix typos before push - rebase with rewrite and edit is your friend)

Issue #169, opened 2023-05-12
on exit, freetype/__init__.py", line 1233, in __del__ TypeError: 'NoneType' object is not callable


HinTak:
This seems to be a race condition on exit - I often get:

```
Exception ignored in: <function Face.__del__ at 0x7f4e89903a60>
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/freetype/__init__.py", line 1233, in __del__
TypeError: 'NoneType' object is not callable
```

The cure is, curious enough, explicitly doing `del face` at the end of the program. This seems to suggest that, maybe there is a bug somewhere about destructor's calling order, such as the  library or other objects are destructed first (and destroying the face handle).

HinTak:
Argh, I thought this has been discussed before, and it has:
https://github.com/rougier/freetype-py/issues/44

HinTak:
This can be a good task to learn about freetype. This can be investigated by building a debug version of freetype, set FT_DEBUG / FT_TRACE to see when and what freetype receives from freetype-py from the other side. This could fix some of the other issues with other `__del__` too.

carandraug:
I've been seeing reports for this error for a couple of years now MicronOxford/cockpit#684 (it's always when the program is exiting so we never paid much attention).

Effectively, the error is at https://github.com/rougier/freetype-py/blob/ccd4f5fca0fb1dd69e40b2966db60139ec1c49da/freetype/__init__.py#LL1231C1-L1231C42 , i.e.:

            FT_Done_Face( self._FT_Face )

And what is `None` is `FT_Done_Face`. I guess this means that `__del__` is being called after ctypes having `dlclose`d the shared library.

carandraug:
I tried to look deeper into this and even `freetype` (the Python module) is already `None` at this point. It appears that it all boils down to [this warning in Python's documentation](https://docs.python.org/3/reference/datamodel.html#object.__del__):

> Warning: Due to the precarious circumstances under which [__del__()](https://docs.python.org/3/reference/datamodel.html#object.__del__) methods are invoked, exceptions that occur during their execution are ignored, and a warning is printed to sys.stderr instead. In particular:
>
[...]

Issue #44, opened 2017-01-26
access violation writing error in Face.__del__()


fhfuih:
Hi, I'm learning to use Freetype recently, and I encounter an exception when I'm using freetype module in python:
```
Exception ignored in: <bound method Face.__del__ of <freetype.Face object at 0x000001AFDAC06860>>
Traceback (most recent call last):
  File "C:\Users\Xiaoqin\AppData\Local\Programs\Python\Python35\lib\site-packages\freetype\__init__.py", line 989, in __del__
OSError: exception: access violation writing 0x0000000000000054
```
I'm running Python 3 on Windows 64-bit and using the latest freetype-py and a homemade 64-bit freetype DLL found in a previous issue post.
I have tried in 2 computers in the same environment above, with a simple single-character rendering program, and have the same exception. Any idea how to deal with it?

rougier:
Could you write a minimal script that show the problem ?

fhfuih:
Sorry for the late reply!
So here's part of a program I wrote:
```
# -*- coding: utf-8 -*-

from freetype import *

class Font(object):
    def __init__(self, filename, size, mono = True):
        self._face = Face(filename)
        self._face.set_pixel_sizes(size, 0)
        self._slot = self._face.glyph
        self._flags = (FT_LOAD_RENDER | FT_LOAD_TARGET_MONO) if mono else FT_LOAD_RENDER
        self.is_mono = mono
        self.name = ' '.join([self._face.family_name.decode('utf-8'), self._face.style_name.decode('utf-8')])

    def load_char(self, char):
        self._face.load_char(char, self._flags)

    def get_data(self):
        return bytes(self._slot.bitmap.buffer)

    def get_size(self):
        if self.is_mono:
            return (self._slot.bitmap.pitch * 8, self._slot.bitmap.rows)
        else:
            return (self._slot.bitmap.width, self._slot.bitmap.rows)
    def get_index(self, charcode):
        return self._face.get_char_index(charcode)

if __name__ == '__main__':
    from PIL import Image
    fnt = Font(r'C:\Windows\Fonts\arial.ttf', 16)
    fnt.load_char(chr(1))
    im = Image.frombytes('1', fnt.get_size(), fnt.get_data())
    im.show()
```

fhfuih:
But your example code works fine. Did a little change:
```
from PIL import Image
from freetype import *

def bits(x):
    data = []
    for i in range(8):
        data.insert(0, int((x & 1) == 1) * 255)
[...]

PR #171, opened 2023-05-18
freetype.Face.__del__: check first if FT_Face_Done has been set to None (#169)

carandraug:
When we're called during the interpreter shutdown, the functions we need may already be set to None.  So check for that before trying to call them which fixes the "'NoneType' object is not callable" errors described in issues #44 and #169.

### DIFF
--- a/freetype/__init__.py
+++ b/freetype/__init__.py
@@ -1227,7 +1227,9 @@ class Face( object ):
         '''
         Discard  face object, as well as all of its child slots and sizes.
         '''
-        if self._FT_Face is not None:
+        # We check FT_Done_Face because by the time we're called it
+        # may already be gone (see #44 and discussion in #169)
+        if FT_Done_Face is not None and self._FT_Face is not None:
             FT_Done_Face( self._FT_Face )
 
 


carandraug:
> Not quite. FT_Done_Face is lib.FT_Done_Face so does not need to be none to break. lib could be none or lib.FT_Done_Face could be none too, in this idea...

`FT_Done_Face` is a *reference* to `lib.FT_Done_Face`. Even if `lib` is set to `None`, `FT_Done_Face` should continue working.

Anyway, on my current case (where this bug is triggered almost always), this check seems to be enough.

carandraug:
I only have one uncertainty about this. The end result is that `FT_Done_Face` will not be called. But that's ok because we're at the stage that the program is exiting anyway (right?).

HinTak:
But in the desired outcome, FT_Done_Face (and FT_Done_Library) should be run, without explicit `del`. To allow multiple faces to used, for example. Otherwise it is no better than do-nothing (and leak face handles) on destruction.

carandraug:
> But in the desired outcome, `FT_Done_Face` (and `FT_Done_Library`) should be run, without explicit `del`. To allow multiple faces to used, for example. Otherwise it is no better than do-nothing (and leak face handles) on destruction.

I think that the desired outcome is to not have memory leaked. This requires calling `FT_Done_Face` (and similar) which is is only available via `Face.__del__`. But in the case that the program is about to exit, then it doesn't matter if those functions are not called because the memory will be freed anyway.

There are multiple situations that need to be "covered":

1. During "normal runtime", `Face.__del__` is called when a `Face` object is about to be destroyed. This is the case where calling `FT_Done_Face` is particularly important to avoid memory leaks. This is working as expected.

2. When the program is about to exit, three different things can happen:

   1. `Face.__del__` may be called early during the interpreter shutdown and have all the functions and variables still available to it. In this situation the current code has no issue and everything is working as expected.
   2. `Face.__del__` may be called later during interpreter shutdown and some variables may already be set to `None`. This is the case that this PR fixes. In this case, `FT_Done_Face` has ben set to `None` and the function is no longer available to be called. But we're about to exit the program so it doesn't matters whether `FT_Done_Face` is actually called because the memory is going to be freed as the program exits.
   5. `Face.__del__` is not called at all. From [Python's documentation itself](https://docs.python.org/3/reference/datamodel.html#object.__del__):
         > It is not guaranteed that [`__del__()`](https://docs.python.org/3/reference/datamodel.html#object.__del__) methods are called for objects that still exist when the interpreter exits.

      In this case, `FT_Done_Face` is never called. As in the case above, it doesn't matter because the program is about to exit and the memory be freed anyway.

   I have seem all these three cases while testing this.

---

The final point is, I think, this: if you want to ensure that `FT_Done_Face` is called when the program exits, you can't rely on `__del__` (but I don't think you need to ensure this).

carandraug:
I'm happy to discuss this further over a video call if you prefer.

HinTak:
The change itself is fine. But I think the problem itself is a bit bigger - namely that freetype-py doesn't currently call FT_Done_X for most X. The most interesting case is perhaps something like a loop which read say, a collection of 100 fonts, one after another, and do stuff with it. Those who care about this situation, at the moment, can do an explicit del Face within the loop (and even that probably does not clean up bitmaps, etc).

I'd say merge this, but keep the issue open.

rougier:
@carandraug Can you open an issue referring to this pull request about the potential problems mentioned by @HinTak?

PR #174, opened 2023-06-29
Ot svg support

HinTak:
It works as is. There is a small problem as I haven't figured out how to populate a c pointer from the python side, so the ctx/state is currently a global instead. If I can't figure it out soon I'll probably add a comment about it and just merge as is.


HinTak:
The global needs to stay at least partially, as otherwise python does not know that the state object kept by the c-side needs to stay around for the next call.

Issue #183, opened 2024-03-08
Please provide non-bundled wheels


zahlman:
I managed to create a pure-Python wheel from source... as desired, it includes only Python source code, and successfully uses the system Freetype shared library via `ctypes`.

However, this process was quite unpleasant. `pip install --no-binary :all: freetype-py` doesn't work like one would hope, at least not with contemporary Pip etc. What happens is that Pip tries to set up build isolation and then also install CMake etc. from source in the isolated environment, which fails. I tried again with `pip wheel --no-binary freetype-py freetype-py` and it works, but it's very slow and unresponsive - it silently downloads and sets up CMake (from a wheel) in the isolated environment.

CMake is a massive dependency - over ten times the size of Setuptools and over a hundred times the size of the resulting installation. It's also completely unnecessary just for packing the `['freetype', 'freetype.ft_enums']` packages into a wheel - `pip wheel` can do that, never mind Setuptools. Even today the download of a wheel like CMake's is non-trivial for some.

As [wheels are preferred even for pure-Python projects](https://pradyunsg.me/blog/2022/12/31/wheels-are-faster-pure-python/), I'd like to request publishing a wheel with just the Python source code needed to talk to the system FreeType. I could do it myself but I wouldn't want to step on this project's toes. It's fine if it uses a completely different PyPI name to avoid user error.

HinTak:
When did anything needing CMake? Upstream freetype is still autotool based...

anthrotype:
if you ask pip to install `--no-binary :all:` you're asking to not use any binary wheels at all, which is why it attempts to build everything from source, including the build requirements like cmake. It's preferable to restrict the option to `--no-binary freetype-py`, which you say worked, despite still setting up the temporary isolated environment and unnecessarily installing cmake (from a wheel, not from source this time).
You can also turn off the PEP518 build isolation with this option, it should be quicker to install then (we can probably document this in README):

`pip install --no-build-isolation --no-binary freetype-py freetype-py`

@HinTak yes, Cmake is a build requirement for building the freetype shared library that we include in the freetype-py wheels published on PyPI, when FREETYPEPY_BUNDLE_FT=1 is set (see setup.py and setup-build-freetype.py scripts).

I think there is a way to only set specific build dependencies conditionally for sdist vs wheel, see:
https://setuptools.pypa.io/en/latest/build_meta.html#dynamic-build-dependencies-and-other-build-meta-tweaks

[...]

PR #184, opened 2024-03-08
use custom PEP517 build backend to require cmake conditionally

anthrotype:
Fixes https://github.com/rougier/freetype-py/issues/183

`pip install --no-binary freetype-py freetype-py` will no longer attempt to download and install cmake in pip's temporary build environment, unless FREETYPEPY_BUNDLE_FT environment variable was set in order to compile the bundled FreeType shared library.

This makes the installation of a pure Python freetype-py (using system-wide FreeType library) quicker and doesn't require one to force --no-build-isolation just to prevent cmake from being installed unnecessarily.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
         platforms: all
 
     - name: Build wheels
-      uses: pypa/cibuildwheel@v2.12.3
+      uses: pypa/cibuildwheel@v2.16.5
       env:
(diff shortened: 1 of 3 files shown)


anthrotype:
also note the CI is currently broken (see latest failed build https://github.com/rougier/freetype-py/actions/runs/8204508642), bumping cibuildwheel to latest like this PR is doing also fixes that, one reason more to merge.

madig (approved):

PR #189, opened 2024-04-14
Fix wrong enum name in Face.get_fstype()

NCBM:
The symbol `FT_FSTYPE_XXX` does not exist in `freetype/ft_enums/ft_fstypes.py`, and instead it is named `FT_FSTYPES` in that file.

`Face.get_fstype()` used the (maybe) wrong name and raises an error when calling it.

### DIFF
--- a/freetype/__init__.py
+++ b/freetype/__init__.py
@@ -1619,7 +1619,7 @@ class Face( object ):
         '''
 
         flag = FT_Get_FSType_Flags( self._FT_Face )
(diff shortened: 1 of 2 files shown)


HinTak:
This looks okay - would you consider adding a test into the tests directory for this also?

NCBM:
> This looks okay - would you consider adding a test into the tests directory for this also?这看起来不错 - 您是否会考虑为此在测试目录中添加一个测试？

Added in `smoke_test.py`; what about this?

PR #203, opened 2025-10-21
Fix the behavior of get_kerning to match FreeType's behavior

### LINKED ISSUE #202: The behavior of get_kerning doesn't match the documentation
Hello,
The documentation of the `get_kerning` function of the `freetype.Face` class seems wrong.

It states that:

> left: The index of the left glyph in the kern pair.
> right: The index of the right glyph in the kern pair.

However, as seen in the code, `get_char_index` is called on both `left` and `right`, meaning that to get a valid kerning value, you should rather provide left and right as characters (I got trapped this way and thought it was a freetype2 bug)

The following solutions could be considered:
- Change the content of the function to match the documentation (however this can cause compatibility issues for those who already did a workaround for this function)
- Change the documentation to match the behaviour of the function (however it can make people feel trapped because they will be forced to provide a character code instead of a glyph)
- Do both! (keep the old function with its new documentation, for compatibility reasons and create a new function, **get_kerning_by_index**, or whatever the name, which takes as input glyph indexes and not char codes)

Thank you for considering this issue,
Have a nice day

ExplodingBottle:
This pull request fixes #202 by changing the behavior of `Face.get_kerning` to match the one of FreeType (this is done by removing the calls to `get_char_index`) and also updates the examples accordingly.

The documentation has been left untouched.

### DIFF
--- a/examples/hello-vf.py
+++ b/examples/hello-vf.py
@@ -27,7 +27,7 @@ if __name__ == '__main__':
         height = max(height,
                      bitmap.rows + max(0,-(slot.bitmap_top-bitmap.rows)))
         baseline = max(baseline, max(0,-(slot.bitmap_top-bitmap.rows)))
-        kerning = face.get_kerning(previous, c)
+        kerning = face.get_kerning(face.get_char_index(previous), face.get_char_index(c))
         width += (slot.advance.x >> 6) + (kerning.x >> 6)
         previous = c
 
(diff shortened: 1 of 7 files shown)


HinTak:
I am somewhat against merging this, but I shall give a better answer when I look at the code in detail.

thisismypassport:
My 2 cents - instead, fix the documentation and put it on your documentation site since https://freetype-py.readthedocs.io/en/latest/api.html seems to currently link to a bunch of empty pages.
The documentation can clarify that unlike the freetype function, the freetype-py function takes chars instead, which is more convenient. (And fixing the documentation isn't a breaking change)

PR #204, opened 2025-11-11
Remove redundant parameter of get_next_char

the-be-to:
The parameter `agindex` in `get_next_char` is not used


HinTak:
While I agree with you this API is poorly implemented - the parameter is a "pass by reference" OUTPUT parameter, changing API with a long history seems Risky.

the-be-to:
> While I agree with you this API is poorly implemented - the parameter is a "pass by reference" OUTPUT parameter, changing API with a long history seems Risky.

In the current implementation `agindex` is immediately reassigned so it doesn't even function as an output parameter here. Also, it seems inconsistent to have such a parameter in `get_next_char` as `FT_Get_First_Char` takes a similar parameter but its wrapper `get_first_char` abstracts it and returns the value directly.

HinTak:
It seems to be returned as "(return, agindex)", I think? I don't disagree with the idea of the pull, but an API change of a frequently used API needs a bit more thought.

the-be-to:
Both `FT_Get_First_Char` and `FT_Get_Next_Char` return a character code directly and a glyph index via a pointer parameter `agindex`. The wrappers  `get_first_char` and `get_next_char` indeed return a tuple of the two values, so the parameter `agindex` would not be needed in either function, but it is nevertheless present in `get_next_char`, while it is absent in `get_first_char`. Aside from that inconsistency, the first line of `get_next_char` assigns a new `FT_UInt` to `agindex`, so whatever was passed in that parameter is lost and the subsequent call to `FT_Get_Next_Char` modifies a new object, leaving the original argument unchanged.

PR #205, opened 2026-02-01
Fetch Freetype source tarballs from a new host

### LINKED ISSUE #200: Freetype Host doesn't seem to have the freetype repo mirrored anymore.
The default FREETYPE_HOST in setup-build-freetype.py seems to not host the file anymore. https://mirrors.sarata.com/non-gnu/freetype/

HexDecimal:
The old host is still unavailable. This is needed to fix build scripts and CI. Resolves #200

GitLab is picked over GitHub because of the archive format, GitHub only has zips.

`tarfile.open` call modified to use universal compression since there's no `.tar.xz` file available.

Test rexgex updated to understand the new source archive filename.

FreeType 2.13.2 accidentally pinned its CMake version to 3.0 which is no longer supported by CMake. `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` is added as a workaround and must be removed after updating FreeType.

### DIFF
--- a/setup-build-freetype.py
+++ b/setup-build-freetype.py
@@ -25,10 +25,10 @@ import platform
 # Needed for the GitHub Actions macOS CI runner, which appears to come without CAs.
 import certifi
 
-FREETYPE_HOST = "https://mirrors.sarata.com/non-gnu/freetype/"
-FREETYPE_TARBALL = "freetype-2.13.2.tar.xz"
+FREETYPE_HOST = "https://gitlab.freedesktop.org/freetype/freetype/-/archive/VER-2-13-2/"
+FREETYPE_TARBALL = "freetype-VER-2-13-2.tar.bz2"
 FREETYPE_URL = FREETYPE_HOST + FREETYPE_TARBALL
-FREETYPE_SHA256 = "12991c4e55c506dd7f9b765933e62fd2be2e06d421505d7950a132e4f1bb484d"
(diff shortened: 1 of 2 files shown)


HinTak:
Your gitlab url is not a release tar ball but auto-generated ones against a tag. There is a small difference.

HexDecimal:
I wouldn't argue otherwise. The main difference being how the archive file is labeled.

Is there a more important difference I'm missing? This is building from source as far as I can tell.

HexDecimal:
FreeType doesn't do any GitHub releases or GitLab releases. The auto generated archives from tags appeared to be the next most reliable option.

HexDecimal:
I'm willing to swap this to a different host URL if you have something in mind.

HexDecimal:
I checked https://download.savannah.gnu.org/releases/freetype/ ~~but the download was slow so I'm not confident it has the capacity for automated tasks. (could just be me though.)~~ I misunderstood the nature of the mirror. It should be fine.

HexDecimal:
I now replaced this with `download.savannah.gnu.org` so it should be the exact same file as before. (same SHA, same filename)

HexDecimal:
The only thing remaining is the decision to to keep, revert, or erase my changes which allow for handling the auto-generated archives which are no longer downloaded by this PR.

The current host of this PR is FreeType's official meta-mirror: `download.savannah.gnu.org` found at https://freetype.org/download.html
I can't imagine a more authoritative source, other than the repos themselves.

